PR #658, opened 2018-06-06
Extract Openapiplugin model as already done with Javaplugin

MikeSchumacherCapgemini:
Implements

Extraction of the openapiplugin model to it's own project as done with the javaplugin.
Created openapi-plugin-parent which contains the model and the plugin.


cobigen/cobigen-openapiplugin-parent/pom.xml
```diff
@@ -0,0 +1,25 @@
...
+  </parent>
+
+  <modules>
+    <module>cobigen-openapiplugin</module>
+    <module>cobigen-openapiplugin-model</module>
+  </modules>
+  
+  <properties>
+    <skip.deployment>false</skip.deployment>
+  </properties>
```
maybeec: please remove

Jenkinsfile
```diff
@@ -62,6 +62,8 @@ node {
 				root = "cobigen/cobigen-core-parent"
 			} else if (origin_branch == "dev_javaplugin") {
 				root = "cobigen/cobigen-javaplugin-parent"
+			} else if (origin/branch == "dev_openapiplugin") {
```
maybeec: here is the bug!

PR #670, opened 2018-06-19
Implementing modification counter for Angular client

jdiazgon:
Implements

* On the client side of Angular, the `modificationCounter` is now incremented, so that the user is able to modify entities from the database multiple times.

@devonfw/cobigen


jdiazgon:
I manually tested it and it works fine. I'm going to wait for Jenkins to build it without errors, and then I will merge it.

jdiazgon:
Wops, I forgot that master wasn't building before. Therefore I'm going to merge directly without waiting Jenkins to build.

PR #671, opened 2018-06-21
UML templates for CobiGen

jdiazgon:
This pull request contains the needed templates for the generation of the `Entity infrastrucutre` from a class diagram, exported as XMI version 2.1.

I have tested that the two-step generation properly generates the files. The server gets deployed correctly.

Implements

* Relationships: Annotations are correctly addressed. If no multiplicity is defined, then CobiGen generates a One-to-One relationship. In addition, they are bidirectional by default.
* The `@Column` annotation is also generated, for the different attributes of the class.

I have left some templates just in case we use them on the future. For example, generating Angular from a class diagram.

@devonfw/cobigen


jdiazgon:
I have another idea for the left templates. They will be stored in my [branch](https://github.com/jdiazgon/tools-cobigen/tree/dev_templates_uml), just in case we need them for the future.

Now I will close this PR and open another one just with the needed files.

PR #677, opened 2018-06-27
Modified input reader to read proper response/parameter type

### LINKED ISSUE #675: OpenApiplugin InputReader doesn't properly read Parameter data
### Expected behavior
Reading info from an OpenApi specification should put the name of the entity as the type of a response/parameter, if it is an entity reference

### Actual behavior
The type of the entity is read (for example object instead of Table)

### Steps to reproduce (bug) 
Generating from an OpenApi specification and trying to read the type of a parameter/response

MikeSchumacherCapgemini:
Adresses/Fixes #675 .

Implements

* Change in the way response/parameter types are read when they are a reference to an entity
* Modified tests to test these changes

### DIFF
--- a/cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
+++ b/cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
@@ -410,7 +410,7 @@ public class OpenAPIInputReader implements InputReader {
                 }
             }
             try {
-                if (Overlay.isReference(param, "schema")) {
+                if (Overlay.isReference(param, Constants.SCHEMA)) {
                     parameter.setIsEntity(true);
                     parameter.setType(schema.getName());
                 } else {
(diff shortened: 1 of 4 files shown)


cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
```diff
@@ -411,13 +412,15 @@ private PropertyDef extractReferenceProperty(OpenApi3 openApi, Schema componentS
             try {
                 if (Overlay.isReference(param, "schema")) {
```
jdiazgon: Could you maybe add _"schema"_ as a constant? If you have a look at the [Constant](https://github.com/MikeSchumacherCapgemini/tools-cobigen/blob/fca94a2e5e1e00e48fcb5e50049de36d7bc81ecf/cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/util/constants/Constants.java) file, maybe you could add it there.

Therefore the final result would be:

`if (Overlay.isReference(param, Constants.SCHEMA)) {`

PR #685, opened 2018-06-28
Fix YAML syntax for Travis CI testing

iriberri:
Adresses/Fixes broken Travis builds, as in: https://travis-ci.org/devonfw/tools-cobigen/jobs/394900583

It seems that the missing space after the `-` was causing that Travis CI couldn't detect the name of the package correctly, and this made the apt-get install command fail.

This should hopefully fix the issue!

@devonfw/cobigen

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ install: true
 addons:
   apt:
     packages:
-      -metaCity
+      - metaCity
 
 jobs:
   include:


maybeec:
Thanks, but unfortunately, it does not :)

maybeec:
I will merge it anyhow and see whether I can continue with it.

maybeec:
I tried every dist (xenial, precise, trusty) without success.
All of them show the same error like this https://travis-ci.org/devonfw/tools-cobigen/jobs/397941416

maybeec:
@iriberri thanks for you help, it seems, that it failed as of the capital c in metaCity. I corrected it to lower case, and now the job is running.

PR #686, opened 2018-06-29
Added EntityRef to ResponseDef

MikeSchumacherCapgemini:
Implements

* EntityDef field containing properties and name of the referenced entity in ResponseDef
* Code to add Entity and its properties to a response in the input reader


MikeSchumacherCapgemini:
One other possibility was to add the Entity as it already was defined to a ResponseDef, though that would require somehow passing it through multiple method calls. 
My proposition would be to leave it like this with an added comment in the wiki to create an issue that more info should be added to this when someone needs it

maybeec:
If the Entity has been already added to the model, you could simply retrieve it again in java code. Then you would not need to pass it around.

Issue #290, opened 2017-02-28
Value of type array of an annotation parameter not printed correctly


maybeec:
On CobiGen generation with javamerge strategy (see [javamerge strategy documentation]( https://github.com/devonfw/tools-cobigen/wiki/cobigen-javaplugin#merger-extensions) 
and [mergeStrategy configuration documentation](https://github.com/devonfw/tools-cobigen/wiki/cobigen-core_configuration#templates-configuration), nested parameters with value type array of annotation are destroyed.

The array notation for annotation parameter values is denoted by `{ ... }` rather than `[ ... ]` as currently printed by `CustomModelWriter` of the java plugin.  See the example below for more explanation:

. code base
```java
import javax.persistence.NamedAttributeNode;
import javax.persistence.NamedEntityGraph;
import javax.persistence.NamedEntityGraphs;
import javax.persistence.NamedSubgraph;

@NamedEntityGraphs(
    value = { @NamedEntityGraph(name = "xxx",
        attributeNodes = { //
            @NamedAttributeNode(value = "xxx", subgraph = "xxx")}, //
        subgraphs = {
            @NamedSubgraph(name = "xxx",
                attributeNodes = @NamedAttributeNode("zzz")), //
            @NamedSubgraph(name = "zzz", attributeNodes = @NamedAttributeNode("aaa")) }) //
    })
public class Whatever {}
```

. code after merge with mergeStrategy javamerge
```java
@NamedEntityGraphs(
    value = { @javax.persistence.NamedEntityGraph(name = "xxx",
        attributeNodes = [ @javax.persistence.NamedAttributeNode(value = "xxx", subgraph = "xxx")],
        subgraphs = [  @javax.persistence.NamedSubgraph(name = "xxx",
                attributeNodes = @javax.persistence.NamedAttributeNode("zzz")), 
            @javax.persistence.NamedSubgraph(name = "zzz", attributeNodes = @javax.persistence.NamedAttributeNode("aaa")) ])
    })
public class Whatever {}
```

What we want to focus on in this bug report is the fact, that the value of type array of `attributeNodes` as well as `subgraphs` is printed with the wrong syntax. Please see that this is not the case for the `value` parameter of this outer annotation, which is kind of strange.

1. [ ] implement a unit test to reproduce the error
2. [ ] fix it

[...]

PR #697, opened 2018-07-10
Added merging annotations to javamerge functionality

### LINKED ISSUE #430: Javamerge does not merge annotations
Currently, the java merger does not merge annotations.
So far there are two restriction after implementation of this ticket:
1. Annotations are not merged recursively as it's not easy to identify "matching" annotations within an array
2. [Repeating annotations](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html) defined next to each other could not be identified / differentiated for a proper merge, therefore, we need to overwrite / ignore all annotations of a given scope as a block

*Acceptance Criteria*
* [x] Annotations, which are not yet there should be added
* [x] All annotations should be replaced or not replaced
* [x] Annotations should be merged the same for fields, methods, classes

MikeSchumacherCapgemini:
Adresses #430.

Implements

* Test case for annotation merging
* Additional method for merging annotations
* updated/deleted old comments/javadoc

### DIFF
--- a/cobigen/cobigen-javaplugin-parent/cobigen-javaplugin/src/main/java/com/devonfw/cobigen/javaplugin/merger/JavaMerger.java
+++ b/cobigen/cobigen-javaplugin-parent/cobigen-javaplugin/src/main/java/com/devonfw/cobigen/javaplugin/merger/JavaMerger.java
@@ -11,6 +11,7 @@ import com.devonfw.cobigen.api.exception.MergeException;
 import com.devonfw.cobigen.api.extension.Merger;
 import com.devonfw.cobigen.javaplugin.inputreader.JavaParserUtil;
 import com.devonfw.cobigen.javaplugin.merger.libextension.ModifyableJavaClass;
+import com.thoughtworks.qdox.model.JavaAnnotatedElement;
 import com.thoughtworks.qdox.model.JavaAnnotation;
 import com.thoughtworks.qdox.model.JavaClass;
 import com.thoughtworks.qdox.model.JavaConstructor;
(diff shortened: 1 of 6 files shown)


maybeec:
Please also come up with a test of more complex annotations like described in #290.
What is the actual behavior there?

jdiazgon:
From my point of view, seems fine.

MikeSchumacherCapgemini:
Complex annotation test is a bit awkward but qdox does not support a way to get further data about anything within an AnnotationValue after getProperty() so checking the string value was the only way i found how this can be tested.

maybeec:
Two things here:

1. Annotations, should be left as is, when it comes to type declarations. In your test, the annotations are after generation printed with full qualified names, while beforehand they had just been printed with simple types. Try to keep the original declaration. I think Qdox also has an interface for this. There are multiple getters for types.
2. Annotations here are just overwritten entirely or not at all. What about recursively merging arrays of 
annotations?
3. Be careful, in Java8 annotations can be repeated: https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html So just comparing as of type of the annotation, you cannot start merging. You have to check whether the annotation is repeatable.

All this issues need to define heuristics how to compare annotations. The easiest way would be to compare the "value", but most often if there is a annotation array, the annotations come with clearly distinguishable values. We can for example introduce something like "if an annotation has 'name' or 'key' as an attribute, the values of such attributes are used as identifier (fallback 'value' value)". What do you think? This is up to discussion.

maybeec:
closed in favor of  #1053

Issue #704, opened 2018-07-17
Refactor template project structure


MikeSchumacherCapgemini:
To make using some maven operations easier as well as possibly avoid performance issues in the future, the structure of the template projects should be improved.

PR #707, opened 2018-07-18
Improve template structure

MikeSchumacherCapgemini:
Adresses #704.

Implements

* src/main/templates resource folder to contain the templates
* added src/main/templates to build path in poms


cobigen-templates/templates-oasp4js/pom.xml
```diff
@@ -36,4 +36,12 @@
...
     </dependency>
   </dependencies>
+  
+  <build>
+    <resources>
+      <resource>
+        <directory>src/main/templates</directory>
+      </resource>
+    </resources>
+  </build>
```
maybeec: please do it just once in the parent pom

cobigen-templates/pom.xml
```diff
@@ -31,6 +31,7 @@
     <resources>
       <resource>
         <directory>${basedir}</directory>
+        <directory>src/main/templates</directory>
         <excludes>
```
maybeec: To have a valid configuration now, you should remove basedir from the resource configurations to not double the sources in the jar and remove the exclusions as they are not needed anymore. Please check the jar build by mvn clean install (in the target folder)

MikeSchumacherCapgemini:
Template configuration not read correctly at the moment, template path is not read from the location of context.xml but from the root path, will try to change it so it looks relative to the context.xml file.
Util classes read relative to templates, will try to change it so they are read from the root package.

maybeec:
Tests are failing. See pr-merge build

Issue #427, opened 2017-09-19
Implement smoketests for all templates


maybeec:
We should have a smoketest for all templates assuring, that there are no exceptions on generation of each template.

Richter-cg:
Tests exist & green:
- [ ] ToInput
- [ ] OpenApiIput
- [ ] XML
- [ ] EtoInput
- [ ] RestServiceInput      (fails on my machine)
- [ ] EntityInput

jdiazgon:
Actually, I saw that you are using an UML xml file as input for the tests. We decided to only use the templates related to the `Entity infrastructure` increment:

- ApplicationPersistenceEntity
- ApplicationEntity
- {variables.className}Entity
- {variables.className}

The rest will be discarded because we are trying to implement a multi-step generation. See [this project](https://github.com/devonfw/tools-cobigen/projects/8) and [this issue](https://github.com/devonfw/tools-cobigen/issues/676).

maybeec:
We have at least smoke tests meanwhile generating all templates within the templates project to assure that there is at least no error on initial generation.

Issue #676, opened 2018-06-26
CobiGen pipeline for reducing templates


jdiazgon:
Currently, one of our major issues is that we have to maintain several plug-ins which make use of similar templates.

For example, the OpenAPI plug-in contains increments/templates like _TOs_ or _CRUD REST services_ which are very similar to the ones used in the Java plug-in.

We could only have templates for the _Entity infrastructure_ and then, using the generated entities as input, generate with them the rest of increments.

Therefore, we must implement some kind of logic inside the CobiGen core to create like a **generation pipeline** for being able to generate in multiple steps.

jdiazgon:
As discussed with @maybeec we have devised some ideas that I would like to reflect:

1. We will add a new configuration feature on the _context.xml_ so that the user is able define a generation pipeline. For instance:

```xml
<gen-pipeline>
	<incrementRef ref="crud_openapi_angular_client_app"/>
	<incrementRef ref="crud_java_server_app::inc21" use-input="${variables.entityName}Entity.java"/>
  </gen-pipline>
```
This configuration means that from an OpenAPI file we will generate the `crud_openapi_angular_client_app` templates. Then, using the generated `${variables.entityName}Entity.java`, we will generate templates from the `crud_java_server_app` using the increment with id _"inc21"_.

2. The second increment will completely wait until the first increment has been generated. Then we need some logic to try to compile the generated files from the first increment (maybe by a method `tryCompile()` to try to compile with each plug-in?). This way, we will be able to use the `pojo.classobject` and reflection. If compiling was not possible, we will have to use the java source file rather than the class object and generate with that data. 

Although this issue is tricky, it will highly improve maintainability of the templates and reduce the development time for the next plug-ins templates.

maybeec:
it should be 
```xml
<gen-pipeline>
	<incrementRef ref="crud_java_server_app"/>
	<incrementRef ref="crud_openapi_angular_client_app::inc21" use-input="${variables.entityName}Entity.java"/>
  </gen-pipline>
```
for a correct example, as the entity is generated by crud_java_server_app.

jdiazgon:
You are right, I misspelled it. I meant generating from an OpenAPI file the entity infrastructure, and then with `crud_java_server_app` generating for instance the REST services.

I have already edited my comment.

jdiazgon:
### Description

With our new CobiGen refactoring, where we will facilitate the creation of new plugins for very different languages, we will have a lot of new templates coming.

[...]

PR #708, opened 2018-07-18
Added XML test.

Richter-cg:
#427 
Added XML generation test.


jdiazgon:
Actually, I saw that you are using an UML xml file as input for the tests. We decided to only use the templates related to the `Entity infrastructure` increment:

- ApplicationPersistenceEntity
- ApplicationEntity
- {variables.className}Entity
- {variables.className}

The rest will be discarded because we are trying to implement a multi-step generation. See [this project](https://github.com/devonfw/tools-cobigen/projects/8) and [this issue](https://github.com/devonfw/tools-cobigen/issues/676).

Maybe you should change your input file.

Issue #678, opened 2018-06-27
Adding the possibility to reference increments from outside the current templates file


jdiazgon:
### Our objective

We would like to add a new feature to reference increments from outside the current _templates.xml_ file.

Currently, we are able to reference increments inside the same file using `incrementRef`. See example [here](https://github.com/devonfw/tools-cobigen/blob/master/cobigen-templates/templates-oasp4j/crud_openapi_java_server_app/templates.xml):

```xml 
<increment name="server" description="CRUD OASP4J Server">
    <incrementRef ref="daos"/>
</increment>
    ....
<increment name="daos" description="DAO's">
    ....
</increment>
```

It would be nice to be able to reference an increment from another templates folder, with a syntax similar to this one:
```xml 
<increment name="server" description="CRUD OASP4J Server">
    <incrementRef ref="crud_java_server_app::rest_service_impl"/>
</increment>
```
The first string of the ref is the name of the templates folder from where to get the increment. After the `::` we state the increment id to reference.

### Related Issues
Part of issue #676.

jdiazgon:
### Current status

From my last PR, the current status of this functionality is the following:

- It is possible to reference any increment from any external trigger. The way to do so is defining an `incrementref` like: 

```xml 
<increment name="server" description="CRUD OASP4J Server">
    <incrementRef ref="crud_java_server_app::rest_service_impl"/>
</increment>
```

The ref string is split using as delimiter `::`. The first part of the string, is the trigger id `crud_java_server_app` to reference. That trigger contains an increment id `rest_service_impl`.

- Although it is possible to reference any trigger, the generation of templates will only work when both triggers match the input file. For instance, when using as input a _yaml_ file, we cannot generate templates that need a Java entity as input file. That logic will be implemented with issues #679 and #680.

### Logic

To implement this functionality, I added the following logic:

1. In [TemplatesConfigurationReader](https://github.com/devonfw/tools-cobigen/pull/717/files#diff-05097276788a210e215d1ebba9c0b0efR542) we read the external incrementref and it is added to the list of increment dependencies.
2. Then we call the [ConfigurationHolder](https://github.com/devonfw/tools-cobigen/pull/717/files#diff-ff3b23382767f377b5955f7359899662R80), which stores the TemplatesConfiguration of the external trigger (the trigger containing the increment we are referencing to). That class contains all the needed information of the trigger like, increments, templates, path to the templates...
3. After that, CobiGen automatically detects the needed templates and is able to resolve their path.

[...]

PR #717, opened 2018-08-02
Reference increments from outside the current templates folder

jdiazgon:
Adresses #678. As stated on the issue, we are implementing a way to reference increments that are not present in the current templates folder (they are defined in another templates.xml file). 

Implements

* A way to reference external increments.
* Some tests to assert this functionality

A SWT Bot test will be added to the `cobigen-eclipse` plug-in after this merge has been completed.

@devonfw/cobigen


jdiazgon:
`ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project core-api: Fatal error compiling: invalid target release: 1.8`

The Java version set on the [Jenkins file](https://github.com/devonfw/tools-cobigen/blob/master/Jenkinsfile#L36) is Java8 and the [master pom](https://github.com/devonfw/tools-cobigen/blob/master/pom.xml#L17) is also set to Java8.

I don't get what is going on.

maybeec:
Thanks, did you adapt the documentation accordingly?

jdiazgon:
Not yet sorry, will do.

PR #725, opened 2018-08-03
Properly load util classes in maven plugin

MikeSchumacherCapgemini:
Adresses #704 .
Dependent on #726 

Implements

* Dependency to latest core version in pom
* modified util class reading so the correct root path is used even when the context.xml file is only found in src/main/templates

### DIFF
--- a/cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
+++ b/cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
@@ -221,7 +221,8 @@ public class GenerateMojo extends AbstractMojo {
                 getLog().error("Could not add configuration folder " + configurationFolder.toString(), e);
             }
         }
-        Path rootPath = null;
+
+        Path rootPath;
         URL rootUrl = classRealm.getResource("context.xml");
         if (rootUrl == null) {
             rootUrl = classRealm.getResource("src/main/templates/context.xml");
@@ -231,6 +232,8 @@ public class GenerateMojo extends AbstractMojo {
             } else {
                 rootPath = Paths.get(URI.create(rootUrl.toString())).getParent().getParent().getParent();
             }
+        } else {
+            rootPath = Paths.get(URI.create(rootUrl.toString()));
         }
         getLog().debug("Found context.xml @ " + rootUrl.toString());
         final List<String> foundClasses = new LinkedList<>();
--- a/cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
+++ b/cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
@@ -133,7 +133,6 @@ public class GenerateMojo extends AbstractMojo {
             getLog().info("");
             return;
         }
-
         List<GenerableArtifact> generableArtifacts = collectIncrements(cobiGen, inputs);
         generableArtifacts.addAll(collectTemplates(cobiGen, inputs));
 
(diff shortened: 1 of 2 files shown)


cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
```diff
@@ -223,18 +222,26 @@ private CobiGen createCobiGenInstance() throws MojoExecutionException, MojoFailu
...
         }
 
-        URL contextXmlUrl = classRealm.getResource("context.xml");
-        if (contextXmlUrl == null) {
-            getLog().error("No context.xml could be found in the classpath!");
-            return null;
+        Path rootPath;
+        URL rootUrl = classRealm.getResource("context.xml");
+        if (rootUrl == null) {
+            rootUrl = classRealm.getResource("src/main/templates/context.xml");
```
maybeec: Just from what I now understood. I am a little bit afraid, that we introduced anoter issue quite some time ago.
I am pretty sure, that copying the template folder to a different place under control, deleting all target and bin folders, and setting this folder as the templateConfigurationFolder in a maven execution, maven will fail generating code in case of any JavaUtil function is invoked from a template.
Can you try it out? 

This is due to the fact, that we rely on an external compiler like eclipse to compile the sources. Such that a maven correct and compliant way would be to call the maven compiler and resource processing on the template folder as well. This would even have resolved the issue with the context.xml as the context.xml will be copied to the target folder at packaging time.

As this feature is not really used by anybody, but it is quite nice for testing, I would like to keep it as is. However, we should be aware of the drawbacks and document them properly. Can you adapt the documentation to that configuration parameter to not let users strugle with this again?

cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
```diff
@@ -223,18 +222,26 @@ private CobiGen createCobiGenInstance() throws MojoExecutionException, MojoFailu
...
+            if (rootUrl == null) {
+                getLog().error("No context.xml could be found in the classpath!");
+                return null;
```
maybeec: Better throw a MojoFailureException and abort processing as not finding the context.xml would result in strange behavior as one of our main assumptions seems to be not valid anymore.

cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
```diff
@@ -272,15 +279,16 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) throws IOExce
...
             getLog().debug("Searching for classes ...");
             final List<Path> foundPaths = new LinkedList<>();
             try {
-                Files.walkFileTree(configFolder, new SimpleFileVisitor<Path>() {
-
+                Files.walkFileTree(rootPath, new SimpleFileVisitor<Path>() {
                     @Override
                     public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                        // LOG.warn("####################");
+                        // LOG.warn("{}", file.toString());
```
maybeec: remove debugging code

cobigen-maven/cobigen-maven-plugin/src/main/java/com/devonfw/cobigen/maven/GenerateMojo.java
```diff
@@ -223,18 +222,26 @@ private CobiGen createCobiGenInstance() throws MojoExecutionException, MojoFailu
...
+            }
+        } else {
+            rootPath = Paths.get(URI.create(rootUrl.toString()));
         }
-        getLog().debug("Found context.xml @ " + contextXmlUrl.toString());
+        getLog().debug("Found context.xml @ " + rootUrl.toString());
         final List<String> foundClasses = new LinkedList<>();
-        if (contextXmlUrl.toString().startsWith("jar")) {
-            getLog().info("Processing configuration archive " + contextXmlUrl.toString());
+        if (rootUrl.toString().startsWith("jar")) {
```
maybeec: its hard to read.
`rootUrl` seems to be the context.xml location, why not naming like this?
`rootPath` is however having different intentions. If the templates are in `src/main/templates` it is pointing to the top level folder root of the configuration. If the templates are located in the root, `rootPath` seems to point at the context.xml Oo

MikeSchumacherCapgemini: Changed the names for easier reading

For clarification:
rootPath should always point at the template project root, for example `.../cobigen-templates-parent/templates-oasp4j`
Upon initilizaiton, if the context configuration is in the rootpackage, it will indeed point to context.xml first, though later on rootPath.getParent() is called to get the actual folder instead. This call is made at this point, as the other way, if the context configuration is in the src/main/templates, you have to somehow get back up 3 folders and getParent() is called thrice on rootPath, after which you have `.../cobigen-templates-parent/templates-oasp4j/src` as you are also starting from the context.xml itself. 
That way, in both occasions, after the next call of getParent() in line 282 you'll get the project root.

If the given input is a jar, the root package is gotten in a different way altogether, splitting the path at the location of the jar indicator `!`, which will be present in both cases too but should not have to be called anyway as the template jar should not have a src/main/templates if i remember correctly

MikeSchumacherCapgemini: Still not done in an elegant way but the easiest way i found to do it without code redundancy.

PR #726, opened 2018-08-03
Make reader properly read from src/main/templates

MikeSchumacherCapgemini:
Adresses #704.
Dependent on #725

Implements

* Added reading from src/main/templates to TemplatesConfigurationReader
* Changed constructor of TemplatesConfigurationReader, as checking the src/main/templates folder for templates.xml was very complicated otherwise, modified tests accordingly
* Added method to get the path of context.xml from ContextConfigurationReader, used in ContextConfiguration to get the correct root path of the templates


cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/generator/GenerationProcessorImpl.java
```diff
@@ -130,7 +130,14 @@ public GenerationReportTo generate(Object input, List<? extends GenerableArtifac
...
         this.forceOverride = forceOverride;
         this.input = input;
+        LOG.warn("{}", logicClasses != null);
         if (logicClasses != null) {
+            LOG.warn("{}", logicClasses.isEmpty());
+            for (Class<?> clazz : logicClasses) {
+                LOG.warn("##############################");
+                LOG.warn("{}", clazz.getSimpleName());
+            }
+            LOG.warn("##############################");
```
maybeec: log output? remove

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/TemplatesConfigurationReader.java
```diff
@@ -92,14 +92,32 @@
...
      * @param templatesRoot
-     *            root path for the template configuration and templates
+     *            root path for the templates
+     * @param templatesName
+     *            name of the folder containing the configuration and templates
      * @throws InvalidConfigurationException
      *             if the configuration is not valid against its xsd specification
      */
-    public TemplatesConfigurationReader(Path templatesRoot) throws InvalidConfigurationException {
+    public TemplatesConfigurationReader(Path templatesRoot, String templatesName) throws InvalidConfigurationException {
```
maybeec: I am a little bit confused. The second (added) parameter is called "templatesName", but it actually is the relative path of the templates folder. Why did you change the API here? using a string for a path is highly not recommended. The sub path could even be constants/named_queries

cobigen/cobigen-core-parent/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/constants/ConfigurationConstants.java
```diff
@@ -19,5 +19,6 @@
     /** Filename of the {@link Properties} used to customize cobigen properties and template relocation. */
     public static final String COBIGEN_PROPERTIES = "cobigen.properties";
 
-    public static final String TEMPLATE_RESOURCE_FOLDER = "src/main/templates";
+    /** Source folder containing templates */
+    public static final String TEMPLATE_SOURCE_FOLDER = "src/main/templates";
```
maybeec: It's a resource folder ;) Source folders are just folder, which compilable resources.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/TemplatesConfigurationReader.java
MikeSchumacherCapgemini: The parameter names should have been changed to projectRoot and templatesFolder or something similar, i simply forgot to do that.

Regarding the API, the problem is that when using the old version, if you have the templates in the new templates source folder, you'll have src/main/templates in the middle of the path, with the project root before and the template folder name after that.
Should the reader not find the configuraiotn in the rootpackage, i couldn't find a decent way of changing the path to look in src/main/templates, as no informaiton about the project root is passed and thus you don't know where exactly to put src/main/templates into the path.
That's why i changed the API, so you get the project root as the first parameter and you can simply resolve the project root with the template folder path, afterwards if the templates.xml is not found there, you resolve the project root with src/main/templates and that result with the template folder path.
Resolving the project root with the template folder path was done in every initilization of the TemplatesConfigurationReader too, which was another reason for me to move it here

cobigen/cobigen-core-parent/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/constants/ConfigurationConstants.java
MikeSchumacherCapgemini: I thought that templates are technically compiled, that's why i changed it

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/TemplatesConfigurationReader.java
maybeec: Ah ok, this is a valid point. However, I would recommend to also set the second parameter to be of type Path. Then you can resolve the path snippets against each other. Please mention in the javadocs, that the first should be absolute, while the second is a relative one.

cobigen/cobigen-core-parent/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/constants/ConfigurationConstants.java
maybeec: No, this notion of source and resource is coming from maven standard. It relates to the compile step in maven, which does not effect the template in any sense.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/TemplatesConfigurationReader.java
MikeSchumacherCapgemini: The resolve() method of the path API actually already covers String as a parameter (https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.lang.String))
When changing the parameter to a path, you'd have to call Paths.get(String), as you take the info from a trigger as a String or manually in tests.
So i'm pretty sure this should be fine, except for the parameter names

maybeec: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#resolve(java.nio.file.Path) ;)

MikeSchumacherCapgemini: Yeah but if it already covers string, i think it would be unnecessary work to invoke Paths.get on the string you'd pass to the implementation i made to pass a path instead, when it is very likely that something similar is already done in the Path api when calling resolve(String)
I couldn't find the actual source but i'm pretty sure the resolve(String) method would look similar to this: 
public Path resolve(String toResolve){
    return resolve(Paths.get(toResolve));
}

Issue #715, opened 2018-08-01
ProviderMismatchException during maven build with Cobigen


tonymodica:
### Expected behavior

As a developer using Cobigen, I want to generate template code from an OpenApi spec file with the Cobigen maven plugin.

### Actual behavior

An exception is thrown and the maven build fails, see appended [maven.log](https://github.com/devonfw/tools-cobigen/files/2249381/maven.log) for details.

[ERROR] Failed to execute goal com.devonfw.cobigen:maven-plugin:4.0.0:generate (generate) on project apigendemo: Execution generate of goal com.devonfw.cobigen:maven-plugin:4.0.0:generate failed: Unable to invoke public abstract java.util.List<com.devonfw.cobigen.api.to.IncrementTo> com.devonfw.cobigen.api.ConfigurationInterpreter.getMatchingIncrements(java.lang.Object) throws com.devonfw.cobigen.api.exception.InvalidConfigurationException. InvocationTargetException: ProviderMismatchException -> [Help 1]

### Steps to reproduce

1. Create a sample OASP4J project. 
2. Put the sample [devonfw.yml.rename.txt](https://github.com/devonfw/tools-cobigen/files/2249379/devonfw.yml.rename.txt) as devonfw.yml in its top level directory.
3. Put the provided [pom.xml.rename.txt](https://github.com/devonfw/tools-cobigen/files/2249378/pom.xml.rename.txt) as pom.xml.
OR use the zipped project [apigendemo.zip](https://github.com/devonfw/tools-cobigen/files/2249374/apigendemo.zip)
4. Call maven package.

I noticed that the same error occurs with templates-oasp4j versions down to 2.4.1 and with current maven 3.5.4 invoked from cli.
(With templates-oasp4j 2.4.0, I get a different error that indicates a bug in the template files, but this is not of concern here)

jdiazgon:
I'm going to finish ASAP my current task, I will investigate it and then I will inform you whether I can do it before my vacations.

jdiazgon:
Some things I found out:

1. On Jenkins, all the Maven plug-in tests pass successfully. However, on my local machine there are two tests failing, which are the only ones that contain the openapi plug-in as dependency. `testPackageInputDataaccessGeneration` and `testEntityInputDataaccessGeneration`.

2. I tried to do `mvn package` using as input a java file. It worked properly and all the templates were generated.

I will keep investigating what is exactly going on here.

PR #727, opened 2018-08-03
Maven build error #715

jdiazgon:
Adresses #715 .

This PR does not contain the solution to the bug stated on the issue. It just contains a test that I used to successfully reproduce the error.

Some things we found out:

* The bug arises when using templates-oasp4j version `2.4.1` onwards.
* The triggers are correctly read.
* When using a Java entity file as input, the generation is successful.

@devonfw/cobigen


MikeSchumacherCapgemini:
If a test should be included for this, it has to be modified a bit, otherwise this pull request can be closed

PR #741, opened 2018-08-15
Reference increments from outside the current templates folder

MikeSchumacherCapgemini:
Adresses #678. Based on #717 

As stated on the issue, we are implementing a way to reference increments and templates that are not present in the current templates folder but instead are defined in another templates.xml file. 
For this to work, the referenced templates folder must be in the same location as the origin, as the reference is parsed using the context.xml.

Implements

* Functionality to reference external increments and templates.
* Tests to assert this functionality and it's exceptions.

An SWT Bot test will be added to the `cobigen-eclipse` plug-in after this merge has been completed.


cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
```diff
@@ -45,13 +57,46 @@ public TemplatesConfiguration readTemplatesConfiguration(Trigger trigger) {
...
+     * the external increment we want.
+     * @param trigger
+     *            to get matcher declarations from
+     * @param incrementToSearch
+     *            name of the increment to search
+     * @return the {@link TemplatesConfiguration}
+     * @throws InvalidConfigurationException
+     *             if the configuration is not valid
+     */
+    public TemplatesConfiguration readTemplatesConfiguration(Trigger trigger, String incrementToSearch) {
```
maybeec: It seems, that it is just possible to reference increments. Can you check that?
What about templateScans and templates to be referenced?
Maybe the code architecture has to rethought to not add multiple parameters per case.

MikeSchumacherCapgemini: According to the documentaiton @jdiazgon created, this is the only thing that's possible so far, i'll have a look into the others as well

MikeSchumacherCapgemini:
templateScanRef has been omitted for now, as it seems complicated to implement as well as not having any apparent uses so far.
In case of a customer or us needing that feature, it can be implemented at a later date.

Issue #749, opened 2018-09-03
QA on templates


maybeec:
### Expected behavior

The current ide setup of devon 2.4.0 delivers templates. I am importing these templates and trying to generate different java application based on different architectures (simple vs complex).
It should work in any combination like 

* CRUD DAO + CRUD logic (simple) + CRUD service REST
* CRUD DAO + CRUD logic (complex) + CRUD service REST
* CRUD DAO + CRUD logic (simple) + CRUD service SOAP
* CRUD DAO + CRUD logic (complex) + CRUD service SOAP
* ... etc

Please also check "nested" and spring data increments. Maybe we should cleanup here. Currently, it is not easy to understand what is meant by "nested".

### Actual behavior

* The annotation @UseCase is not defined anymore. Most probably, it is removed in oasp4j? @hohwille 
* In some cases, Cto methods are missing on logic layer, but they are generated and accessed through the service implementation.
* I think, I also observed, that any AbstractUseCase or some super class was not generated in some combination. 

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Generate different combinations as mentioned above. Different errors will arise.
2. Please start with a clean state all the time.

### Related/Dependent Issues
- 

### Implementation Hints
Best would be, if we could provide multiple SWTBot tests, each generating one combination of increments above, which should compile out of the box. Then we could check that, by trying to find compilation errors in the error view. You might want to have a look at the latest commit on cobigen-eclipse, which introduced SWTBot code to access the error log. A prove of validity might be to assert, that there are no errors after generation and rebuild.

jdiazgon:
The usefulness of nested and spring data increments should be discussed. I haven't received any comment on my side related to compilation errors. This may mean that our developers just use the most common increments like `CRUD DAO + CRUD logic (simple) + CRUD service REST + CRUD service SOAP + TO's + Entity infrastructure`.

PR #750, opened 2018-09-06
#749 fixing CRUD logic complex increment

jdiazgon:
Addresses #749.

This pull request just tries to fix the compilation errors related to `CRUD Logic (with use cases)` increment. It does not include any SWT Bot test. However, I would like to quickly fix this by publishing the template changes on a new accumulative patch for devonfw.

**Implements**

Increment `CRUD Logic (with use cases)` had a compilation error related to `${entityName}Cto` not defining `find${entityName}Cto` method. I have added this method into UcFind${variables.entityName}Impl.java template because I thought it made sense to have it on this use case as it is called "Find".

@devonfw/cobigen

### DIFF
--- a/cobigen-templates/templates-oasp4j/src/main/templates/crud_java_server_app_complex/templates/java/${variables.rootPackage}/${variables.component}/logic/api/usecase/UcFind${variables.entityName}.java.ftl
+++ b/cobigen-templates/templates-oasp4j/src/main/templates/crud_java_server_app_complex/templates/java/${variables.rootPackage}/${variables.component}/logic/api/usecase/UcFind${variables.entityName}.java.ftl
@@ -25,4 +25,19 @@ public interface UcFind${variables.entityName} {
    */
   PaginatedListTo<${variables.entityName}Eto> find${variables.entityName}Etos(${variables.entityName}SearchCriteriaTo criteria);
 
+  /**
+   * Returns a composite ${variables.entityName} by its id 'id'.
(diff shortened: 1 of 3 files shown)


maybeec:
Generating Cto's should be a separate increment. Maybe this caused some more confusion in th epast as it has not been implemented like this for every option.
The reason: Cto's can be generated but especially in complex projects generated CTO's come with a lot of objects. Most often, this is nonsense as you want to customize your CTO's based on the current use case and optimize it in case of data to be submitted over the link. Generating CTO's by default would force such kind of projects to delete them afterwards. So we should have them seperately to the TO's.

jdiazgon:
> Generating Cto's should be a separate increment. So we should have them separately to the TO's.

I have some questions about this:
- If you generate the `CRUD Logic (all in one)` increment, it is included the `${variables.component#cap_first}Impl.java` template which defines methods to find ETOs and CTOs. Would I have to remove the methods related to CTOs? Otherwise, the user would have to select TOs and CTOs increments anyway for not getting compilation errors.

So in my opinion, separating the increments would confuse even more our developers, because if you generate and get compilation errors because of CTOs not being found, then a normal developer would think "Huh, I need to select CTO increment too then or either remove the methods". So anyway he will end up having to make changes in his code.

jdiazgon:
Closing this PR as now there have been a lot of templates changes and it does not make sense to change the old templates.

Issue #435, opened 2017-09-26
make PermissionConstants overriding


maybeec:
Currently, you have to reselect the generation of PermissionConstants in oasp4j templates if the target already exists. Since we now have the ability to do this on the level of configuration, we should do it.

PR #756, opened 2018-09-19
#435 checked-in for make permission constant overriding

sanjaykumarcg:
I am creating again this pull request due to problem with xml setting in eclipse .
Adresses #435.

changes 
Add java merge strategy in template.xml file for making permission constant overridable .

@devonfw/cobigen


maybeec:
@jdiazgon sorry that's not right, permission contants have to be in sync with the xml definition. So it should overwrite the complete file normally.

PR #761, opened 2018-09-25
#738 deleted embeddables templates

### LINKED ISSUE #738: Embeddables should not generate interface and transfer-object
When I use @Embeddable objects/classes in entities and use CobiGen, then the generated code will have this features:
* an interface and ETO is generated for the entity (by entity infrastructure) and all will reference the embedable class in its fields, getters and setters
* dozer will map all these properties so the embeddable will be copied and I will not have the reference to the JPA mapped embeddable object in persistent state. This is what we want for data sovereignty.

However, CobiGen supports generation with @Embeddable class as input (if in `persistence.api` and Suffix is `Embaddable`). This will generate an interface and a transferobject (ETO) also for the embeddable. Even when this is done, the ETO for the entity using the embeddable will reference the Embeddable directly. Also there is no easy way to abstract from the embeddable way and there is no need to do so (we should follow KISS and this interface and ETO for the embeddable is IMHO just pointless).
Finally the embeddable should not be located in persistence layer if it is directly referenced from a transferobject. Even though the embeddable has a reference to JPA it should therefore considered as part of the API and transfer-objects and go to the same location.

My suggestion is therefore:
* simply remove the embeddable templates (or give some rationale for them to explain what I am missing)
* improve the documentation (IMHO we should create a separate issue in oasp4j/devon4j and have it documented in the JPA guide in embeddable section).

sanjaykumarcg:
Remove templates of embeddables /Fixes #738.

I have done changes for Embeddables should not generate interface and transfer object .
I have removed embeddables templates folder and remove trigger of embeddables from context.xml .

@devonfw/cobigen


maybeec:
@jdiazgon @sanjaykumarcg have you checked, that generation works for embeddables? I have seen a lot of exceptional logic for embeddables also in oasp4j server templates. Please check!

Issue #778, opened 2018-10-25
Built jar of templates-oasp4j is incorrect


jdiazgon:
### Expected behavior

When doing a `mvn clean package` on templates-oasp4j I would expect the jar file to contain the templates on the correct path _src/main/templates_

### Actual behavior

The jar file is created storing the templates on the root folder as shown below:

![image](https://user-images.githubusercontent.com/31688036/47493965-deb22400-d850-11e8-8115-706b268171d0.png)

Which is incorrect as the templates structure has changed.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Do `mvn clean package -Dmaven.test.skip=true` on the _\CobiGen-templates_ folder.

PR #779, opened 2018-10-25
Built jar of templates-oasp4j is incorrect

jdiazgon:
Adresses #778 .

Changed the pom so that when the jar is created, the templates folder are correctly stored on `src/main/templates` instead of the root folder.

@devonfw/cobigen


jdiazgon:
I should test whether this new structure of the jars works fine with the CobiGen tests.

PR #785, opened 2018-11-08
Refreshing relative project

jdiazgon:
This PR implements a way to refresh a relative project. It is useful when a template is going to be relocated to a different project from the original one (the one where the input is located). Therefore the path is relative (i.e. projectname-core/../api/src/main...) and we need to get that relative project (i.e. projectname-api).

It only does work in case of having a project on the system path like this:

projectname /
├── core /
│   ├── InputFile.java
├── api/
│   ├── template will be generated here
└── 

It is very hard to make this implementation generic for every project, as they may have different structures.

@devonfw/cobigen

### DIFF
--- a/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/common/tools/PathUtil.java
+++ b/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/common/tools/PathUtil.java
@@ -1,6 +1,5 @@
 package com.devonfw.cobigen.eclipse.common.tools;
 
-import java.io.File;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashSet;
@@ -8,7 +7,6 @@ import java.util.Set;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.ResourcesPlugin;
 
 /**
  * The {@link PathUtil} class provides some common functionality on workspace relative paths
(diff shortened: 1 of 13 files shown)


jdiazgon:
**Attention:** This PR is included in PR #790

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/common/tools/PathUtil.java
```diff
@@ -90,4 +92,32 @@ public static String getProject(String path) {
...
+
+            // We want to get the project name after "/../" which is the child project
+            int nextChildProject = filePath.indexOf("/../") + 4;
+            filePath = filePath.substring(nextChildProject, filePath.length());
+            filePath = projectName + "-" + filePath;
+
+            return ResourcesPlugin.getWorkspace().getRoot().getProject(PathUtil.getProject(filePath));
+        }
+        return project;
+    }
```
maybeec: This method seems much to specific to a certain maven project hierarchy. It could even be a totally different project out of scope of `../`.
Please have a look at https://github.com/devonfw/tools-cobigen/blob/99ba18b5bc589c04ef8623f7fbd9684c320aeca0/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java#L557 This function should do what you want to do in a more generic way. Maybe you can move this function and reuse it in both locations.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/wizard/generate/control/GenerateSelectionJob.java
```diff
@@ -84,8 +84,11 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException, Inte
                     cobigenWrapper.getWorkspaceDependentTemplateDestinationPath(generationReport.getGeneratedFiles());
                 Set<IProject> projects = Sets.newHashSet();
                 for (String filePath : generatedFiles) {
+
                     IProject project =
                         ResourcesPlugin.getWorkspace().getRoot().getProject(PathUtil.getProject(filePath));
+                    project = PathUtil.getRelativeProjectIfNeeded(filePath, project);
```
maybeec: you are overwriting the project variable all the time? Wouldn't that make just sense if the project is null or does not exist?

maybeec:
looks good, anyhow, there is a lot in this pr, which is not relating to the feature implemented, what about the other stuff? is it accidentially here from another PR? if so, when can we merge it?

jdiazgon:
> looks good, anyhow, there is a lot in this pr, which is not relating to the feature implemented, what about the other stuff? is it accidentially here from another PR? if so, when can we merge it?

So I added those commits because I thought that it could cause some kind of conflict on the future if I did not add them. It seems it was not a good idea, probably it would be easier if I create a new PR with the needed commits. What do you think?

jdiazgon:
Closing as it contains some not related commits.

Issue #788, opened 2018-11-14
No feedback when updating templates without connection


jdiazgon:
### Expected behavior

As a developer, I want to know if there was any connectivity problem while trying to download the templates so that I am aware that the templates were not downloaded or that I do not have connection

### Actual behavior

If templates are not downloaded/updated, I do not get any message stating that information.

### Steps to reproduce

1. Disconnect from the internet.
2. Click on `Update templates`
3, You won't get any feedback.

PR #787, opened 2018-11-14
Cobigen eclipse update templates feature

sanjaykumarcg:
Fixes #766.

i have checked-in HealthCheckDialog.java for check NULL pointer exception 

@devonfw/cobigen

PR #781, opened 2018-11-05
Dev eclipseplugin

sanjaykumarcg:
Adresses/Fixes #684 and #740 .

Implements

1. Adapt the CobiGen Template and this  option will ask user to adapt the templates which was downloaded from "Update templates" inside CobiGen_Templates project. User should agree to overwrite the changes done by him inside templates. Once he agrees, we will update the latest templates inside the project
2. i have made changes in GeneratorWrapperFactory.java for resolving when we trying CobiGen generate template  it was throwing error .
3. i have remove the isUpdateDialogShown flag check and check only generate project name is exist or not .

@devonfw/cobigen

PR #790, opened 2018-11-16
Improving feedback to the user when updating templates

jdiazgon:
Adresses #788.

Implements

* When you do not have internet connection, you receive some feedback about it.
* When adapting templates, you receive feedback that the templates are being imported
* Fixed an issue that forced to download the latest templates each time you clicked on `generate`.

**Attention:** This PR also includes PR #785 and PR #787 

@devonfw/cobigen


maybeec:
I am not sure, how this relates to #787 as it contains the same commits, please have a look at the review of #787.
Even #781 seems to be the same. Which PR now contains the latest changes? Please close the others.

jdiazgon:
@maybeec with the last two commits, I have fixed all your change requests but [this one](https://github.com/devonfw/tools-cobigen/pull/787#discussion_r235282215), which I think should be discussed.

Issue #807, opened 2018-12-12
Automatically convert the just imported `CobiGen_Templates` folder to a Maven project


sanjaykumarcg:
# Expected behavior

We want after imported CobiGen_Templates it should be as maven project

Currently it is normal project .

### By grammatically it should be done .
 

Below is some piece of code may be it will be helpful . 

private static final String OCB_PLUGIN_FOLDER = "C://your_folder_with_jars";
File directory = new File();
   //get all the files from a directory
   PrintWriter writer = new PrintWriter("update_repo_maven.bat", "UTF-8");
   writer.println("rem "+ new Date());
   File[] fList = directory.listFiles();
   for (File file : fList){
       if (file.isFile()){
       String absolutePath = file.getAbsolutePath() ;
       Manifest  m = new JarFile(absolutePath).getManifest();
       Attributes attributes = m.getMainAttributes();
       String symbolicName = attributes.getValue("Bundle-SymbolicName");

       if(symbolicName!=null &&symbolicName.contains("com.yourCompany.yourProject")) {
           String[] parts =symbolicName.split("\\.");
           String artifactId = parts[parts.length-1];
           String groupId = symbolicName.substring(0,symbolicName.length()-artifactId.length()-1);
           String version = attributes.getValue("Bundle-Version");
           String mavenLine= "call mvn org.apache.maven.plugins:maven-install-plugin:2.5.1:install-file -Dfile="+ absolutePath+" -DgroupId="+ groupId+" -DartifactId="+ artifactId+" -Dversion="+ version+" -Dpackaging=jar ";
           writer.println(mavenLine);
       }

       }
   }
   writer.close();

sanjaykumarcg:
we can use plugin alsomay be this is also helpful .

<project>
    <groupId>org.test.maventemplate</groupId>
    <artifactId>JarName</artifactId> (A fldernamed JarName was created) 
    <version>1.0-SNAPSHOT</version>
    <packaging>jar</packaging>

    <name>JarName</name>
    <url>http://maven.apache.org</url>

    <properties>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <build>
        <plugins>
            <plugin>
                <groupId>org.jibx</groupId>
                <artifactId>jibx-maven-plugin</artifactId>
                <version>1.2.4</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>bind</goal>
                        </goals>
                    </execution>
                </executions>
[...]

PR #814, opened 2018-12-27
Automatically convert CobiGen_Templates to a Maven project

jdiazgon:
Adresses #807.

This PR implements a way to automatically convert _CobiGen_Templates_ to a Maven project. It works fine, though my implementation is indeed hackish:

I have really tried to find a way to execute the command _configure -> to Maven project_ programmatically, but I did not find any good documentation explaining this.

The way I have implemented the solution is by directly calling the Eclipse Class that contains that logic ([EnableNatureAction](https://github.com/jibx/jibx-eclipse/blob/master/jibx-eclipse-plugin/src/org/jibx/eclipse/builder/EnableNatureAction.java)).

The drawback here is that the class is not API and I get the following warning:

image

If there is a better solution, I would like to hear it. From the Spy plug-in this is the information I have retrieved:

```txt
The active contribution item identifier:
org.eclipse.m2e.enableNatureAction

The active contribution location URI:
menu:org.eclipse.ui.projectConfigure?after=org.eclipse.m2e.enableNatureAction

The active contribution item class:
EnableNatureAction

The contributing plug-in:
org.eclipse.ui.workbench (3.107.1.v20160120-2131)
```

@devonfw/cobigen

### DIFF
--- a/cobigen-eclipse/cobigen-eclipse/META-INF/MANIFEST.MF
+++ b/cobigen-eclipse/cobigen-eclipse/META-INF/MANIFEST.MF
@@ -15,6 +15,7 @@ Require-Bundle: org.eclipse.ui;bundle-version="3.104.0",
  org.eclipse.core.runtime;bundle-version="3.8.0",
  org.eclipse.ui.navigator;bundle-version="3.5.200",
  org.eclipse.ui.navigator.resources;bundle-version="3.4.400",
+ org.eclipse.m2e.core.ui;bundle-version="1.6.0",
  org.eclipse.core.filesystem;bundle-version="1.5.0"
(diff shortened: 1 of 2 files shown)


maybeec:
There is an easier and more compliant way: http://www.vogella.com/tutorials/EclipseProjectNatures/article.html#get-and-set-a-project-natures
After that you have to trigger a refresh and rebuild most probably programmatically as well. The EnableNatureAction most probably does this in one go, but it is not defined as an official API, so the access to it is discouraged.

jdiazgon:
I have tried many solutions but none of them seems to work. By decision of @maybeec and me, we are going to merge this PR.

PR #818, opened 2019-01-21
Improve Cobigen reliese script

sanjaykumarcg:
DEVC-78.

i have dome the changes inside git_repo file for init_submodule method .

@devonfw/cobigen

### DIFF
--- a/scripts/src/tools/git_repo.py
+++ b/scripts/src/tools/git_repo.py
@@ -15,7 +15,7 @@ class GitRepo:
 
     def __init__(self, config: Config, path: str = None) -> None:
         self.__config: Config = config
-
+        
         try:
             if not path:
                 self.__repo = Repo(config.root_path)
@@ -52,6 +52,7 @@ class GitRepo:
     def update_and_clean(self):
         log_info("Executing update and cleanup (git pull origin && git submodule update && git clean -fd)")
         self.pull()
+        self.__repo.git.submodule("init")        
         self.__repo.git.submodule("update")
         self.__repo.git.clean("-fd")
         if not self.is_working_copy_clean():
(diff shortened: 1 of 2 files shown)


scripts/src/tools/git_repo.py
```diff
@@ -61,8 +62,8 @@ def update_and_clean(self):
 
     def checkout(self, branch_name):
         log_info("Checkout " + branch_name)
-        self.__repo.git.checkout(branch_name)
-        self.update_and_clean()
+        self.__repo.git.checkout(branch_name)        
+        #self.update_and_clean()
```
jdiazgon: I don't exactly remember if we commented the `update_and_clean()` for any special reason, but if not, can you check if the script stills works with this line uncommented?

I think it is a good practice to call this method each time there is a checkout.

scripts/src/tools/git_repo.py
```diff
@@ -135,17 +136,14 @@ def merge(self, source: str, target: str) -> None:
                 self.__repo.git.execute("git merge --abort")
                 self.reset()
                 sys.exit()
-    def init_submodule(self, submodule_path: str) -> None:               
-        self.__repo.git.submodule("init") 
-        self.__repo.git.execute("git submodule init")        
-        self.__repo.git.submodule("update")
-                         
+    def init_submodule(self, submodule_path: str) -> None:
```
jdiazgon: Please remove this method as it is not needed now.

jdiazgon:
Please also remove the last 3 commits, as you have [added and modified](https://github.com/devonfw/tools-cobigen/pull/818/commits/50a95e8700f907407d2f57613e8c2851a8b1807f) the submodules like tools-cobigen submodule and devonfw-docgen submodule, but this was not intended.

When the changes are tested I will merge.

scripts/src/tools/git_repo.py
sanjaykumarcg: okay,I am checking with uncommented and let you know the result .

sanjaykumarcg: After uncomment it is still working fine .

PR #832, opened 2019-02-20
TS-merger improvements

jdiazgon:
Addresses issue [30](https://github.com/devonfw/ts-merger/issues/30) and [29](https://github.com/devonfw/ts-merger/issues/29).

Implements on the ts-merger:

* Interfaces get merged correctly.
* Export statements get merged correctly.

The only case that is not working is when the `export` statement does not contain brackets:

`export a from 'b'`.

@devonfw/cobigen


jdiazgon:
Locally it builds correctly. On Jenkins is not building most probably due to memory issues:

`Killed`

Issue #816, opened 2019-01-14
Inter-process communication for CobiGen refactoring


jdiazgon:
For the CobiGen refactoring (take a look into [this document](https://teams.microsoft.com/_#/docx/viewer/teams/https:~2F~2Fcapgemini.sharepoint.com~2Fsites~2FDevonfw~2FShared%20Documents~2FCobigen~2FRequirementsCobiGen.docx?threadId=19:ce387815d8d3403ebf9da3af31d13a94@thread.skype&baseUrl=https:~2F~2Fcapgemini.sharepoint.com~2Fsites~2FDevonfw&fileId=0DDBCADE-9582-4118-8B8A-4C698F774CF0&ctx=files&viewerAction=view) to get more information about it) we need to find the best approach for [Inter-process communication](https://en.wikipedia.org/wiki/Inter-process_communication) as we will spawn different processes in different programming languages that need to communicate between each other.

We will use this issue to discuss the different strategies.
### Approaches:

- **File creation:** Process A creates a file on the disk which contains the parameters and data the process B needs. Process B reads this file and then performs its operations. Finally, B writes another file on disk with the result of the execution that A reads. The simplest way yet the less performant as it means the creation of files on HDD.
- **H2 Database:** Create a local H2 database where we store all the needed information (paths of the files) and the external process is able to read and write to it. Similar to first option but probably more performant. If we had two Eclipse instances, then each of them would create a unique H2 database.
- **[Message queue](https://en.wikipedia.org/wiki/Message_queue)**: It is an asynchronous communications protocol in which a linked list of messages is stored within the kernel and identified by a message queue identifier.
- **[Socket](https://en.wikipedia.org/wiki/Network_socket):** Data sent over a network interface to a different process. If we use TCP protocol we have a big drawback when multiple Eclipse instances are running. If we try to generate with CobiGen on these different instances, we would have blocked ports. We have to find a correct way of acquiring ports randomly. Another option to discuss is **UDP**.

In this issue, @maybeec and me will try to discuss the different approaches and add new ones if needed.

jdiazgon:
> I am very interested in the Message queque approach. Is it really cross-os? Is it easily implementable in different languages?

Nice to know that you are interested on it, it is also my first option to investigate :) 

Wikipedia says that `Most operating systems` use this method. And also there seems to be some tools providing this [cross-os messaque queue](https://stackoverflow.com/a/14036067)

jdiazgon:
[...]

PR #840, opened 2019-03-14
First external plug-in approach

jdiazgon:
First approach of  #816 

Implements:

* External process handling: Contains the ability to execute servers on an specific port and terminate them when requested.
* Provides methods for sending HTTP requests to the server.
* Exception handler for the errors related to the connection with the server.
* Added a dummy test to check whether we are able to run an exe file.

@devonfw/cobigen


cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/exceptions/ConnectionExceptionHandler.java
```diff
@@ -8,19 +8,42 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Exception handler for all the connectivity issues with the external process
+ */
 public class ConnectionExceptionHandler {
```
maybeec: What are all this message fields here about? I am not sure, whether I get the idea here.

jdiazgon: So my idea was to create an exception handler that unified all the different exceptions that can arise when trying to connect to the server (IO exception, MalformedURL, Protocol exception...). 

Therefore, we need to set the different messages depending on the exception that got thrown and in which context was thrown.

By doing it like this, code seems cleaner to me because we don't have 4 `catch` statements on each call to the server. If there is a better way, please tell it to me as I was not sure how to correctly handle exceptions in a clean way.

maybeec: Ok, fine, but the question for me is then, why are the messages not set here by default as constants? Why are there setters and getters for the messages?

Please also have a look at the ExceptionUtil. I think your instanceof checks for getting the cause is not a good idea as every of these exceptions could be hidden as a cause of another exception.
Also one question in regards to the IOExceptions. You are sleeping for a while and then continue. What's the reason? I cannot see any effect on this.

jdiazgon: Changes implemented.

>  You are sleeping for a while and then continue. What's the reason?

Actually, this is because when I execute the external server, I want to wait some time for it to get deployed (executed). That's why I thought it was a good idea to sleep a little bit in each new retry.

jdiazgon:
Finally the JUnit test has been fixed. The problem was that Jenkins did not have permissions to run the exe file.

PR #841, opened 2019-03-14
Killing Nashorn, adding merger on external process

jdiazgon:
Adresses  #816.

**NOTE**: This PR adds the external server (45 MB), in `src/test/resources`. Until I findd a better approach, I create this pull request to get some initial feedback :)

Implements:

* Removing Nashorn from the TypeScript plug-in.
* Connecting to an external server that implements merger and beautifier.
* Changing tests to fit new functionality

@devonfw/cobigen


cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -4,32 +4,55 @@
...
 
+    /** Logger instance. */
+    private static final Logger LOG = LoggerFactory.getLogger(TypeScriptMerger.class);
+
+    /**
+     * Instance that handles all the operations performed to the external server, like initializing the
+     * connection and sending new requests
+     */
+    ExternalProcessHandler request = ExternalProcessHandler
+        .getExternalProcessHandler(ExternalProcessConstants.HOST_NAME, ExternalProcessConstants.PORT);
```
maybeec: Should be private if not used somewhere else

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -4,32 +4,55 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.devonfw.cobigen.api.constants.ExternalProcessConstants;
 import com.devonfw.cobigen.api.exception.MergeException;
 import com.devonfw.cobigen.api.extension.Merger;
+import com.devonfw.cobigen.impl.exceptions.ConnectionExceptionHandler;
+import com.devonfw.cobigen.impl.externalprocess.ExternalProcessHandler;
 import com.devonfw.cobigen.tsplugin.merger.constants.Constants;
+import com.devonfw.cobigen.tsplugin.merger.transferobjects.FileTO;
+import com.devonfw.cobigen.tsplugin.merger.transferobjects.MergeTO;
```
maybeec: As of devonfw architecture the package would be "tos" and the Tos in camel case will be written with large T and small o :)

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
+     * @param osw
+     *            {@link OutputStreamWriter} used for writing the data to be sent
+     * @throws IOException
+     *             when connection to the server failed
+     * @throws JsonGenerationException
+     *             When generating the JSON from the transfer object failed
+     * @throws JsonMappingException
+     *             When mapping the JSON from the transfer object failed
+     */
+    private void sendRequest(Object dataTO, HttpURLConnection conn, OutputStream os, OutputStreamWriter osw)
```
maybeec: dataTo

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
+        throws IOException, JsonGenerationException, JsonMappingException {
+        ObjectWriter objWriter;
+        objWriter = new ObjectMapper().writer().withDefaultPrettyPrinter();
+        String jsonMergerTO = objWriter.writeValueAsString(dataTO);
```
maybeec: jsonMergerTo

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
+        // We need to escape new lines because otherwise our JSON gets corrupted
+        jsonMergerTO = jsonMergerTO.replace("\\n", "\\\\n");
+
+        osw.write(jsonMergerTO);
+        osw.flush();
+        os.close();
+        conn.connect();
+
+        if (conn.getResponseCode() != HttpURLConnection.HTTP_CREATED) {
+            throw new RuntimeException("Failed : HTTP error code : " + conn.getResponseCode());
```
maybeec: Please make use of the CobiGen Exception (CobiGenRuntimeException), this would differentiate exceptions accidentially occurred and the ones we fired explicitly

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
-                    importsAndExports.append(LINE_SEP);
-                } else {
-                    body.append(line);
-                    body.append(LINE_SEP);
-                }
-            }
-        } catch (IOException e) {
-            throw new MergeException(base, "Could not process merged contents for formatting.", e);
-        }
+        FileTO fileTO = new FileTO(body);
```
maybeec: fileTo

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
+
+        StringBuffer bodyBuffer = new StringBuffer();
+        try (OutputStream os = conn.getOutputStream(); OutputStreamWriter osw = new OutputStreamWriter(os, "UTF-8");) {
+
+            sendRequest(fileTO, conn, os, osw);
 
-        String formattedBody =
-            executeJS(base, invocable -> invocable.invokeMethod(((ScriptEngine) invocable).eval("global"),
-                "js_beautify", body.toString()), Constants.BEAUTIFY_JS);
+            BufferedReader br = new BufferedReader(new InputStreamReader((conn.getInputStream())));
```
maybeec: Memory leak :)

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
maybeec: Maybe this is also an exception, we should serve by the cobigen API because if there are connection problems, we may want to implement some retry mechanism in future in cobigen-core. So the exception should be easy to be differentiated.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
+        } catch (IllegalStateException e) {
+
+            LOG.error("Closing connection on InputReader.", e);
+            request.terminateProcessConnection();
+        } catch (Exception e) {
+            connectionExc.setConnectExceptionMessage("Connection to server failed, attempt number " + 0 + ".");
+            connectionExc.setIOExceptionMessage("IO exception when merging");
+
+            connectionExc.handle(e);
+        }
```
maybeec: I think there is a great potential to share the http connection code in a separate method serving just url to be called and the parameters to be passed serving back the result.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
```diff
@@ -63,107 +86,125 @@ public String getType() {
...
-                } else {
-                    body.append(line);
-                    body.append(LINE_SEP);
-                }
-            }
-        } catch (IOException e) {
-            throw new MergeException(base, "Could not process merged contents for formatting.", e);
-        }
+        FileTO fileTO = new FileTO(body);
+        HttpURLConnection conn = request.getConnection("POST", "Content-Type", "application/json", "beautify");
```
maybeec: Isn't this conn a closable as well? I am not quite sure. If so, we have an open handle here which is not closed.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
maybeec: btw. maybe even code to be shared in cobigen API? as every plugin would need this code right?

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
maybeec: 2. is "beautify" the URL? If so, we should start wit naming conventions to not have any name clashes later on. So here it could be tsplugin/beautify

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/constants/Constants.java
```diff
@@ -21,6 +21,11 @@
     /** Export statement regex */
     public static final String EXPORT_REGEX = "export\\s+([^\\s]+)";
 
+    /**
+     * Path of the executable file where the server of the external process is stored
+     */
+    public static final String EXE_PATH = "src/test/resources/testdata/externalserver/nestserver.exe";
```
maybeec: you are accessing the test resources from the src main java. This will not work after packaging.
In addition, src/test/resources and src/main/java folders will vanish at build time and their contents will be placed on root.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/constants/Constants.java
```diff
@@ -21,6 +21,11 @@
...
+
     /** We want to check whether it is a real export statement, not like the type "export class a" */
     public static final Map<String, Boolean> NOT_EXPORT_TYPES;
```
maybeec: btw. is this needed anymore? The whole merging is done in ts or is there any logic also in java?

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
jdiazgon: I have created this HTTP exception handling [here](https://github.com/devonfw/tools-cobigen/pull/840/files#diff-45a2daf5ecab0802bb8ac0f498aa8a99). Can you check if this is what you meant?

It returns true when the HTTP request was 200 or 201, and false otherwise.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
jdiazgon: I agree, you will find it implemented [here](https://github.com/devonfw/tools-cobigen/pull/840/files#diff-3c4d8d9150d5465bd1104dbbf44849b4R310)

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
jdiazgon: I think that I have implemented something similar with:

`HttpURLConnection conn = request.getConnection("POST", "Content-Type", "application/json", "beautify");`

`getConnection(...)` is a method I implemented [here](https://github.com/devonfw/tools-cobigen/pull/840/files#diff-3c4d8d9150d5465bd1104dbbf44849b4R271).

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
jdiazgon: `The resource type HttpURLConnection does not implement java.lang.AutoCloseable` 

This is something you already suggested me on a previous pull request 😄. Unfortunately, it is not closeable.

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/constants/Constants.java
maybeec: but wouldn't this be solved if you make the beautify also on the language server?

jdiazgon: The beautify process is already done on the server side, however it still messes up the imports. For instance, from this:

`import {bla, ble} from bla/path`

It converts it to:

```TypeScript
import {
bla,
ble
}
from
bla/path
```

Splitting the imports from the body could be done on the server side. However, I don't see the point of doing it like that, as we have it already implemented on the Java side + we reduce the data to send over IPC..

cobigen/cobigen-tsplugin/src/main/java/com/devonfw/cobigen/tsplugin/merger/TypeScriptMerger.java
maybeec: No, I just meant, that on any download error, we might want to serve a secific type of exception, which we can handle to make retries. Therefore CobiGenRuntimeException might be too generic to not retry NPE for example.

jdiazgon:
Closing because I have opened a clean pull request #862

PR #842, opened 2019-03-15
devonfw/tools-cobigen#DEVC-255: Update CobiGen ionic templates

jambulud:
devonfw/tools-cobigen#DEVC-255.

Implements

* Update OpenApi ionic templates to Ionic 4
* Update Eto ionic templates to Ionic 4

@devonfw/cobigen


jdiazgon:
Thanks for your PR @jambulud, I'm going to check whether everything is working as expected.

cobigen-templates/templates-devon4j/src/main/templates/crud_ionic_client_app/templates/src/app/pages/${variables.etoName#lower_case}-list/${variables.etoName#lower_case}-list.page.scss.ftl
```diff
@@ -37,12 +37,12 @@
 }
 
 .selected {
-    --background: var(--color-light);
-    --color: var(--colors-dark);
+    --background: #cbdeff;
+    --color: #222;
```
maybeec: is there any reason why moving from variables to actual values?
@jdiazgon Do we use scss here? I am not sure.

Normally, we should prefer variables for css if supported (e.g. by scss). Otherwise, maintainability of generated code will decrease.

jambulud: The reason is that those variables where added in the "variables.scss" file which is generated by default on ionic v4 projects. In order to add the used variables it is necessary to add that .scss file as template, however, if having an extra template "variables.scss.ftl" is not a problem, those variables would be added again.

jdiazgon: I agree with Malte, I would add that extra template, it is worth having it.

PR #847, opened 2019-03-22
Html merger added <html> tags

jdiazgon:
Html merger added some extra tags that currently break the latest version of Ionic 4. It added to the html file:

```html
<html>
  <head></head>
  <body>
    <rest of the Ionic file>
  </body>
</html>
```
This produced strange errors, that's why we have decided to remove those extra tags. I have tested it in both Ionic and Angular and it is working fine.

@devonfw/cobigen


jdiazgon:
I'm going to merge it as I need this for releasing the latest version of Ionic 4.

PR #855, opened 2019-03-29
Initial cobigen CLI commit

sanjaykumarcg:
Fixes #DEVC197.

Implements
1. I have created TestPicocli.java as main class and i have created one more class CreateJarFile.java as Utils .
2. i have created Junit Test case TestPicocliTest.java and CreateJarFileTest.java file .
3. i have restore Entity file in src\test\resources\testdata\unittest\EmployeeEntity.java 

@devonfw/cobigen


sanjaykumarcg:
Please review this pull request

cobigen-cli/.classpath
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdiazgon: This `.classpath` file should be added to the `.gitignore` list, so that it doesn't get pushed to the repo :)

cobigen-cli/src/main/java/cobigen.bat
```diff
@@ -0,0 +1,21 @@
+ECHO OFF
+set path=C:\Program Files\Java\jdk1.8.0_181\bin
+set CLASSPATH=.;C:\Program Files\Java\jdk1.8.0_181;C:\Program Files\Java\jdk1.8.0_181\lib\dt.jar;C:\MyData\IDE4\workspaces\com.cobigen.picocli\picocli-2.2.1.jar;C:\Users\syadav9\.m2\repository\com\devonfw\cobigen\openapiplugin\2.2.0\openapiplugin-2.2.0.jar
```
jdiazgon: We need to find a way to use paths that does not depend on your computer, like relative paths. 

Also, for the Java installation, most probably we need something similar to what the devonfw-distribution does. If you go to the file `/scripts/environment-project-bat` you will see something like this:

```
set JAVA_HOME=%SOFTWARE_PATH%\java
```

However, I'm not pretty sure about this solution. What do you think @maybeec ?

cobigen-cli/src/main/java/cobigen.bat
```diff
@@ -0,0 +1,21 @@
...
+ECHO "Welcome to cobigen and below are increment available for your input" 
+
+    
+set arg0=%0
+set arg1=%1
+set arg2=%2 
+javac TestPicocli.java
+set /p answer= $cg -g increment=%arg1%
+
+java -jar C:\MyData\IDE4\workspaces\com.cobigen.picocli\src\main\java\cobigen.jar
```
jdiazgon: Also here we need to use relative paths ;)

.metadata/.plugins/org.eclipse.core.runtime/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,2 @@
+eclipse.preferences.version=1
```
jdiazgon: You should add `.metadata` folder to the `.gitignore` file. We don't need all these files.

cobigen-cli/src/main/java/cobigen.bat
```diff
@@ -34,7 +34,7 @@ REM echo setting PATH
 REM setx PATH %JAVA_HOME%\bin;%PATH% -m
 REM echo Display java version
 REM java -version
-set CLASSPATH=.;C:\Program Files\Java\jdk1.8.0_181;C:\Program Files\Java\jdk1.8.0_181\lib\dt.jar;C:\MyData\IDE4\workspaces\com.cobigen.picocli\picocli-2.2.1.jar;C:\Users\syadav9\.m2\repository\com\devonfw\cobigen\openapiplugin\2.2.0\openapiplugin-2.2.0.jar
+set CLASSPATH=.;%JAVA_HOME%;C:\Program Files\Java\jdk1.8.0_181\lib\dt.jar;C:\MyData\IDE4\workspaces\com.cobigen.picocli\picocli-2.2.1.jar;C:\Users\syadav9\.m2\repository\com\devonfw\cobigen\openapiplugin\2.2.0\openapiplugin-2.2.0.jar
```
jdiazgon: Thanks for the change! However, I still see paths that depend on your computer, like `C:\MyData\IDE4\workspaces\com.cobigen.picocli\picocli-2.2.1.jar` or `C:\Users\syadav9\.m2\repository\com\devonfw\cobigen\openapiplugin\2.2.0\openapiplugin-2.2.0.jar`. There we should use relative paths.

jdiazgon:
Closing as this PR contains too many commits with not needed files. I will create a fresh one.

PR #862, opened 2019-04-12
Killing Nashorn, adding merger on external process

jdiazgon:
Adresses  #816.

**NOTE**: Replacing pull request #841 because it added binaries of 45 MB.

Implements:

* Removing Nashorn from the TypeScript plug-in.
* Connecting to an external server that implements merger and beautifier.
* Changing tests to fit new functionality

@devonfw/cobigen


jdiazgon:
Before merging, I need to fix the Jenkins build. Most probably is not working because yet I have not added binaries for Linux.

Issue #158, opened 2015-10-30
language agnostic templates


hohwille:
Note: This is a visionary feature request that will cause a lot of effort and refactoring. However, I think that it would cause a huge benefit.

It would be awesome if the templates could be compatible with the language they produce. The template-engine could have a grammar configuration for each language that defines how the language syntax is used to express templating logic.
E.g. for Java this would mean:
- Within the code you could have macros like for loops, etc. in comments (see comments below)
- variables in package and classnames are expressed via some silly syntax like `x_varname_x` instead of `${varname}` what would also be allowed in package segments, type names, methods, fields, etc. 

UPDATE: initially proposed `$_varname_$` or `__varname__` but due to strange problems with tools (eclipse, git) we changed to prevent problems.

The new variable syntax with underscores could also express the case-transformation (assuming variable names are then treated case-insensitive):
* `x_varname_x` for lowercase
* `X_VarName_X` for PascalCase
* `x_varName_x` for camlCase
* `X_VARNAME_X` for UPPERCASE
* `X_VAR_NAME_X` for CAP_CASE
* `x_var-name_x` for train-case

(Please note that I am intentionally mixing concepts of velocity and freemarker, because I think both have pros and cons but we need to create something new and better).
- The most tricky part is IMHO to allow writing control statements that are not comments and can be compiled as Java but allow dynamic usage of the templating-engine:
- If you have method cascades such as currently `${var}.getFoo().getBar()` and the type of `${var}` is also dynamic it can get really tricky. However I assume that `getFoo()` has to assume at least a common supertype to work on and then you could also do

```
Supertype /* <#replace-left varType> */ $_var_$ = ...;
$_var_$.getFoo().getBar();
```

If the type is easy to extend you could also create a subclass X_VarType_X in the CobiGen templates that extends from Supertype and mark it as not being a template so the class itself will be ignored on generation.
It would be nice to make some experiments and prototyping in this area to get more insights.
The same concept would easily work with JS, .NET, PHP, python, ruby, scala, etc.
Using this strange v_var_v approach might look ugly but it would be a very simple way to solve big problems.
Assume you can refactor your CobiGen Templates with Java in a typesafe way...

What I like very much about velocity and miss a lot in freemarker is direct access to java calls:
- `foo.bar.some` for `foo.getBar().getSome()`
- `foo.myMethod()` for `foo.myMethod()`
[...]

PR #879, opened 2019-05-14
Fixed stupid replacement bug

hohwille:
Seems that the new variable syntax was never really used. As I am working on #158 I was hitting this stupid bug that does not remap the temporary character back ending up with crazy outcome.


hohwille:
Replaced by #880

PR #892, opened 2019-05-19
Issue865 related bugfix

### LINKED ISSUE #865: Angular services increment can be improved
### Description

A month ago we released a new set of templates that enabled the possibility to generate REST services on Angular, using as input a yaml file.

This increment is called **CRUD devon4ng Service Based Angular** which contains two child increments:

image

The issue is that to be able to generate services that compile, you need to select also the _model_ increment, which is part of **CRUD devon4ng Angular app**:

image

One of our users has suggested to change this, as it is confusing. We should duplicate that _model_ increment to **CRUD devon4ng Service Based Angular**.

1. Paste the following yaml file into the root of your project

```yml
openapi: 3.0.0
servers:
  - url: "https://localhost:8081/server/services/rest"
    description: Just some data
info:
  title: devonfw Example
  description: Example of an API definition
  version: 1.0.0
  x-rootpackage: com.devonfw.poc.jwtsample
paths:
  /authormanagement/v1/author/{authorId}:
    get:
      operationId: findAuthor
      parameters:
        - name: authorId
          in: path
          required: true
          schema:
            type: integer
            format: int64
            minimum: 0
            maximum: 50
      responses:
        "200":
          description: Any
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Author"
            text/plain:
              schema:
                type: string
        "404":
          description: Not found
components:
  schemas:
    Author:
      x-component: authormanagement
      description: Entity definiton of Author
      type: object
      properties:
        authorId:
          type: integer
          format: int64
          uniqueItems: true
        name:
          type: string
          maxLength: 30
          minLength: 3
          uniqueItems: true
        address:
          type: string
          maxLength: 100
          minLength: 3
          uniqueItems: true
        email:
          type: string
          maxLength: 100
          minLength: 3
          uniqueItems: true
      required:
        - authorId
  requestBodies:
    AuthorData:
      content:
        application/json:
          schema:
            $ref: "#/components/schemas/Author"
      required: true

```

2. Click on _CobiGen -> Generate_.

mandybuchhold:
Adresses/Fixes #865.

Implements

* is related to issue 865 and fixed a bug that occurred after adding a increment

@devonfw/cobigen

### DIFF
--- a/cobigen-eclipse/cobigen-eclipse-test/.classpath
+++ b/cobigen-eclipse/cobigen-eclipse-test/.classpath
@@ -5,6 +5,7 @@
 	<classpathentry kind="src" path="src/main/java/"/>
 	<classpathentry exported="true" kind="lib" path="lib/slf4j-api-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/assertj-core-3.8.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-test-5.1.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/jcl-over-slf4j-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/log4j-over-slf4j-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/logback-classic-1.1.2.jar"/>
--- a/cobigen-eclipse/cobigen-eclipse-test/.classpath
+++ b/cobigen-eclipse/cobigen-eclipse-test/.classpath
@@ -5,7 +5,6 @@
 	<classpathentry kind="src" path="src/main/java/"/>
 	<classpathentry exported="true" kind="lib" path="lib/slf4j-api-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/assertj-core-3.8.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/core-test-5.1.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/jcl-over-slf4j-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/log4j-over-slf4j-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/logback-classic-1.1.2.jar"/>
--- a/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java
+++ b/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java
@@ -544,14 +544,14 @@ public abstract class CobiGenWrapper extends AbstractCobiGenWrapper {
      *         <code>false</code> otherwise
      */
     private boolean isMergableFile(Set<TemplateTo> templates) {
-        if (templates != null) {
-            for (TemplateTo template : templates) {
-                if (template.getMergeStrategy() != null
-                    && !template.getMergeStrategy().equals(ConfigurationConstants.MERGE_STRATEGY_OVERRIDE)) {
-                    return true;
-                }
+
+        for (TemplateTo template : templates) {
+            if (template.getMergeStrategy() != null
+                && !template.getMergeStrategy().equals(ConfigurationConstants.MERGE_STRATEGY_OVERRIDE)) {
+                return true;
             }
         }
+
         return false;
     }
 
@@ -585,14 +585,14 @@ public abstract class CobiGenWrapper extends AbstractCobiGenWrapper {
      *         <code>false</code> otherwise
      */
     private boolean isOverridableFile(Set<TemplateTo> templates) {
-        if (templates != null) {
-            for (TemplateTo template : templates) {
-                if (template.getMergeStrategy() != null
-                    && template.getMergeStrategy().equals(ConfigurationConstants.MERGE_STRATEGY_OVERRIDE)) {
-                    return true;
-                }
+        // if (templates != null) {
+        for (TemplateTo template : templates) {
+            if (template.getMergeStrategy() != null
+                && template.getMergeStrategy().equals(ConfigurationConstants.MERGE_STRATEGY_OVERRIDE)) {
+                return true;
             }
         }
+        // }
         return false;
     }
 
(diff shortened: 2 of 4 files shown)


cobigen-eclipse/cobigen-eclipse-test/.classpath
```diff
@@ -5,6 +5,7 @@
 	<classpathentry kind="src" path="src/main/java/"/>
 	<classpathentry exported="true" kind="lib" path="lib/slf4j-api-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/assertj-core-3.8.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-test-5.1.0.jar"/>
```
jdiazgon: Do you know why this jar has been added to your classpath?

mandybuchhold: I don't know. I saw i had an error in cobigen-eclipse-test in Jenkins and in my local eclipse there was this classpath additional and i tried to add this also and see if this makes some difference. Normaly I changed there nothing, therfore i didn't commit it first.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java
```diff
@@ -544,10 +544,12 @@ public boolean isMergableFile(String path, Collection<IncrementTo> consideredInc
      *         <code>false</code> otherwise
      */
     private boolean isMergableFile(Set<TemplateTo> templates) {
-        for (TemplateTo template : templates) {
-            if (template.getMergeStrategy() != null
-                && !template.getMergeStrategy().equals(ConfigurationConstants.MERGE_STRATEGY_OVERRIDE)) {
-                return true;
+        if (templates != null) {
```
maybeec: I am not totally sure, whether this solves the issue correctly.
I would guess you got a null pointer exception while referring to an increment of another trigger, which @jdiazgon initially implemented. For any reason, the list of templates is empty, which for me seems to be more of an invalid state of the application. Maybe the templates of the referenced increment has been tried to accessed and could not determined. Just checking not null here would mean, that you will not determine whether a path in the preview is going to be merged or not. Most probably by your fix, every path resulting from a externally referenced increment is shown as non-mergable, although it might be.

jdiazgon: > I am not totally sure, whether this solves the issue correctly.

You are right. Actually this was my idea, to simply add the `if` condition as a workaround. The thing is that Mandy and me tried to debug this but we were stuck, so I decided that maybe for this kind of small issues, we should not spent too much time.

The thing is that before adding the `if`, a window with a NPE was popping up when trying to generate code. But this NPE was not harmful at all because anyway it generated all the templates.

maybeec: You are right. It will generate, but the merge / generate preview will not show the truth.
So for me its a more severe problem as we are playing with the trust of the people to use cobigen.

jdiazgon: True. When I have time, I will take a look into this.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/wizard/common/control/CheckStateListener.java
```diff
@@ -236,46 +236,6 @@ private void checkPageComplete() {
...
-     * simultaneously, parents will be unselected if at least one child is not selected, and parents will be
-     * automatically selected if all children of the parent are selected
-     * @param event
-     *            triggering {@link CheckStateChangedEvent}
-     * @param packageSelector
-     *            current {@link CheckboxTreeViewer} for the package selection
-     */
-    private void performCheckLogic(CheckStateChangedEvent event, CheckboxTreeViewer packageSelector) {
-
-        SelectIncrementContentProvider cp = (SelectIncrementContentProvider) packageSelector.getContentProvider();
```
jdiazgon: In my opinion, we should not move this method to `SelectFileLabelProvider` because that class is only related to labels. Also the JavaDoc of `CheckStateListener` specifies the following:

> This CheckStateListener provides the check / uncheck of the increments list and the generationresource tree and synchronizes both in order to get a consistent view

This method should stay here.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java
```diff
@@ -582,12 +576,14 @@ public boolean isOverridableFile(String path, Collection<IncrementTo> considered
      *         <code>false</code> otherwise
      */
     private boolean isOverridableFile(Set<TemplateTo> templates) {
+        // if (templates != null) {
```
jdiazgon: Please remove this comment and the one below :)

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/wizard/common/control/CheckStateListener.java
```diff
@@ -245,32 +247,52 @@ private void checkPageComplete() {
...
-                boolean allChecked = true;
-                for (Object child : cp.getChildren(parent)) {
-                    if (!packageSelector.getChecked(parent.createChildPath(child))) {
-                        allChecked = false;
+                if (event.getElement().toString().contains("All")) {
+                    packageSelector.setAllChecked(true);
+                }
+
+                // checks if all child increments are checked and checks All-Checkbox
+                boolean allChecked2 = true;
```
jdiazgon: I wouldn't use variable name with numbers. I have never seen them. Please find a better name, or even use `allChecked`. The boolean from above does not conflict with this one

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/CobiGenWrapper.java
```diff
@@ -564,10 +560,6 @@ private boolean isMergableFile(Set<TemplateTo> templates) {
      */
     public boolean isOverridableFile(String path, Collection<IncrementTo> consideredIncrements) {
         if (path != null) {
-            if (consideredIncrements.contains(new ComparableIncrement(ALL_INCREMENT_ID, ALL_INCREMENT_NAME, null,
```
jdiazgon: I had a call with Mandy to check this, and it seems that in a special case it throws a warning saying that increments got duplicated (it still generates). So she will investigate this further.

Issue #878, opened 2019-05-14
OpenAPI definition of server path is not being correctly generated


jdiazgon:
### Behavior

CobiGen is not able to generate correctly the paths defined on OpenAPI. It does not generate the first segment of the path. For instance:

```YAML
/authormanagement/v1/author/:
    get:
      operationId: findAuthor
      responses:
        '200':
          description: Any
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Author'
            text/plain:
              schema:
                type: string
        '404':
          description: Not found
```

If you use the _Documentation_ increment, there is a part where the URL should be generated like this: `https://localhost:8081/server/services/rest/authormanagement/v1/author/`, but actually it generates:

```java
https://localhost:8081/server/services/rest/Author/v1/author/{authorId}/
```

Because it is retrieving the component schema `Author` from `$ref: '#/components/schemas/Author'`

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Copy the following yaml file to the root folder of your project

```YAML
openapi: 3.0.0
servers:
  - url: 'https://localhost:8081/server/services/rest'
    description: Just some data
info:
  title: Devon Example
  description: Example of a API definition
  version: 1.0.0
  x-rootpackage: com.devonfw.poc.angular
paths:
  /authormanagement/v1/author/:
    get:
      operationId: findAuthor
      responses:
        '200':
          description: Any
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Author'
            text/plain:
              schema:
                type: string
        '404':
          description: Not found
components:
    schemas:
        Author:
          x-component: Author
          description: Entity definiton of Author
          type: object
          properties:
            name:
              type: string
              maxLength: 30
              minLength: 3
              uniqueItems: true
 
```
2. Generate using increment _Documentation_
3. You will see that the generated URL is wrong.

### Related/Dependent Issues

#870 

### Hints

Take a look to the _OpenAPIInputReader_. The fix will be there.

PR #894, opened 2019-05-21
#870 Fixing openapi short paths

### LINKED ISSUE #870: OpenAPI short paths break CobiGen
### Current behavior

CobiGen throws the following error message when paths are shorter than two slashes "/":

image

An example is:
```yaml
paths:
  /authormanagement:
```
If it is like this, then it works fine:
```yaml
  /authormanagement/v1/author/:
```
We should fix this bug which arises on `OpenAPIInputReader extractPaths()` line 424

1. Copy the following OpenAPI definition into the root of your project:

```yaml
openapi: 3.0.0
servers:
  - url: "https://localhost:8081/server/services/rest"
    description: Just some data
info:
  title: devonfw Example
  description: Example of an API definition
  version: 1.0.0
  x-rootpackage: com.devonfw.poc.jwtsample
paths:
  /authormanagement:
    x-component: authormanagement
    post:
      responses:
        "200":
          description: Any
```
2. Click on _CobiGen -> Generate_ and the error will arise.

Mghcapgemini:
Adresses/Fixes #870.

* Changed lazy path extraction with regex based one. 
* Renoved unnecessary condition (if substring is in string)
* Added exception handling for short paths 

@devonfw/cobigen


cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
```diff
@@ -413,33 +415,53 @@ private PropertyDef extractReferenceProperty(OpenApi3 openApi, Schema componentS
...
-                for (int i = 3; i < mp.length; i++) {
-                    pathUri = pathUri.concat(mp[i] + "/");
+
+                try {
+                    match = "^\\/[^\\/]+\\/+[^\\/]+\\/(.+)";
+                    pattern = Pattern.compile(match);
+                    matcher = pattern.matcher(pathKey);
+                    matcher.find();
+                    pathUri += matcher.group(1);
+                    if (pathUri.charAt(pathUri.length() - 1) != "/".charAt(0)) {
```
jdiazgon: `"/".charAt(0)`

Wouldn't be easier to just compare

`"/"`

Therefore the `if` would look like this:

`if (pathUri.substring(pathUri.length() -1) != "/")`

cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
```diff
@@ -413,33 +415,53 @@ private PropertyDef extractReferenceProperty(OpenApi3 openApi, Schema componentS
...
+                        pathUri += "/";
+                    }
+                } catch (IllegalStateException e) {
+                    ;
```
jdiazgon: With the code from above, the Exception is not needed.

cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
```diff
@@ -413,33 +415,53 @@ private PropertyDef extractReferenceProperty(OpenApi3 openApi, Schema componentS
...
-                        if (path.getOperations() == null) {
-                            path.setOperations(new ArrayList<OperationDef>());
-                        }
-                        operation.getParameters()
-                            .addAll(extractParameters(paths.get(pathKey).getOperation(opKey).getParameters(),
-                                paths.get(pathKey).getOperation(opKey).getTags(),
-                                paths.get(pathKey).getOperation(opKey).getRequestBody()));
-                        path.getOperations().add(operation);
+
+                root_component = mp[1];
```
jdiazgon: `root_component` is not being used anywhere. It would be very interesting to add a new parameter on `PathDef`, like the following:

`PathDef path = new PathDef(rootComponent, pathUri, version);`

This would fix this issue #878 

Also remember that Java is Camel Case ;), it should be `rootComponent`.

cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
```diff
@@ -413,33 +415,53 @@ private PropertyDef extractReferenceProperty(OpenApi3 openApi, Schema componentS
...
                 String pathUri = "/";
-                for (int i = 3; i < mp.length; i++) {
-                    pathUri = pathUri.concat(mp[i] + "/");
+
+                try {
+                    match = "^\\/[^\\/]+\\/+[^\\/]+\\/(.+)";
+                    pattern = Pattern.compile(match);
+                    matcher = pattern.matcher(pathKey);
+                    matcher.find();
+                    pathUri += matcher.group(1);
```
jdiazgon: There is a better way to implement this, you can check first whether this group contains at least 2 groups:

```java
if (matcher.groupCount() > 1) {
                    pathUri += matcher.group(1);
                    if (pathUri.substring(pathUri.length() - 1) != "/") {
                        pathUri += "/";
                    }
                }
```

In this case, you don't need to throw any exception.

Mghcapgemini: Actually, I m only extracting one group, which is `(.+)`, the part that comes after **rootComponenet/version/thisPartIsExracted**. If the given path is too short <-> does not match the written match, the exception occurs.
In this case, `matcher.groupCount()` would never be > 1.

cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
Mghcapgemini: Unlike chars, `==` checks the object references. That's why I went for chars. I ll change this to String comparison but with the `equals()` function instead.

jdiazgon: Yeah, you are right, it should be `equals()`, I always forget it 👍

cobigen/cobigen-openapiplugin-parent/cobigen-openapiplugin/src/main/java/com/devonfw/cobigen/openapiplugin/inputreader/OpenAPIInputReader.java
jdiazgon: After discussing this, we found out that `groupCount()` shouldn't be used. We will go for other approaches like `if (matcher.matches())` or `if (matcher.find())`

cobigen/cobigen-openapiplugin/src/main/resources/META-INF/LICENSE.txt
```diff
@@ -1,201 +0,0 @@
-                                 Apache License
```
jdiazgon: Can you add again this file? I think we should not remove it

PR #907, opened 2019-06-11
T716 xml plugin modifications to support heuristic InputReader choice 

### LINKED ISSUE #716: Improve reader performance by heuristically find the right plug-in to read an input
### Expected behavior

As a user of CobiGen, I want to generate code based on a openAPI file without getting any errors in the log like `1:1: Content is not allowed in prolog.`

### Actual behavior

As we are trying to use all available parsers in a row to determine the language used in a file, the XML parser may be used before openAPI causing `1:1: Content is not allowed in prolog.` error messages if the file parsed is not of XML.

1. Take an openAPI file and start generating
2. See the red marked error log as stated above

### Related/Dependent Issues

#713 
#487

### Proposed Solution

Let us implement a heuristic, which enables parser selection based on the file ending. This would even improve performance. Therefore, we should introduce a new method in the InputReader API like 

```java
/**
 * Try to determine, whether the input file is readable by the plug-ins parser. If false is returned by the method, it is not assured, that the read method is not called at all. In case of the situation, that no plug-in is most likely to read the input, CobiGen will try every input reader to read the input. Please be aware, that this method should be kept highly performant. It does not have to be an exact result.
 * @return true, if it is most likely, that the file can be read by this input reader's read method<br>
 * false, if the reader is most likely not able to read the file. 
 */
boolean isMostLikelyReadable(Path path)`
```

The Java8 default implementation should return `false` to not introduce the necessity to adapt every plug-in after changing the `InputReader` interface.

Mghcapgemini:
Adresses/Fixes #716 .

Implements
Heuristic choice of InputReader based on file extension

* Added isMostLikelyReadable implementation to the xml plugin

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-xmlplugin/src/main/java/com/devonfw/cobigen/xmlplugin/inputreader/XmlInputReader.java
+++ b/cobigen/cobigen-xmlplugin/src/main/java/com/devonfw/cobigen/xmlplugin/inputreader/XmlInputReader.java
@@ -256,7 +256,7 @@ public class XmlInputReader implements InputReader {
     @Override
     public boolean isMostLikelyReadable(Path path) {
         List<String> validExtensions = Arrays.asList("xml", "xmi");
-        String fileExtension = FilenameUtils.getExtension(path.toString());
+        String fileExtension = FilenameUtils.getExtension(path.toString()).toLowerCase();
         return validExtensions.contains(fileExtension);
     }
 }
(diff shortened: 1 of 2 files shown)


cobigen/cobigen-xmlplugin/src/main/java/com/devonfw/cobigen/xmlplugin/inputreader/XmlInputReader.java
```diff
@@ -250,4 +251,11 @@ public Object read(Path path, Charset inputCharset, Object... additionalArgument
         }
         throw new IllegalArgumentException("Currently folders are not supported as Input by XmlInputReader#read");
     }
+
+    @Override
+    public boolean isMostLikelyReadable(Path path) {
+        String validExtension = "xml";
```
jdiazgon: Can you also add `xmi`? It is an standard for UML diagrams. CobiGen is able to read a class diagram and generate code from it.

Mghcapgemini: Done

PR #908, opened 2019-06-11
T716 eclipse plugin modifications to support heuristic InputReader choice 

### LINKED ISSUE #716: Improve reader performance by heuristically find the right plug-in to read an input

Mghcapgemini:
Adresses/Fixes #716 .

Implements
Heuristic choice of InputReader based on file extension
* Added isMostLikelyReadable usage in the eclipse plugin
* Removed repetitive code

@devonfw/cobigen

### DIFF
--- a/cobigen-eclipse/cobigen-eclipse/.classpath
+++ b/cobigen-eclipse/cobigen-eclipse/.classpath
@@ -20,7 +20,7 @@
 	<classpathentry exported="true" kind="lib" path="lib/mmm-util-core-7.4.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/slf4j-api-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/javaplugin-model-2.0.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/core-5.2.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/core-api-5.2.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-5.3.0-SNAPSHOT.jar" sourcepath="/core"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-api-5.3.0-SNAPSHOT.jar" sourcepath="/core-api"/>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
--- a/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/generic/FileInputConverter.java
+++ b/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/generic/FileInputConverter.java
@@ -1,6 +1,7 @@
 package com.devonfw.cobigen.eclipse.generator.generic;
 
 import java.nio.charset.Charset;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 
@@ -36,7 +37,6 @@ public class FileInputConverter {
      *             if any exception occurred during converting the inputs or creating the generator
      */
     public static List<Object> convertInput(CobiGen cobigen, List<Object> resources) throws GeneratorCreationException {
-
         List<Object> convertedInputs = Lists.newLinkedList();
         String inputType = null;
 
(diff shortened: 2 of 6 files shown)


cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/generic/FileInputConverter.java
```diff
@@ -51,17 +51,24 @@
...
+                    readAndAddInput(cobigen, convertedInputs, inputType, readerType, inputFilePath, charset);
+                    continue;
+                }
+
+                else if (cobigen.isMostLikelyReadable("openapi", inputFilePath)) {
+                    readerType = "openapi";
+                    readAndAddInput(cobigen, convertedInputs, inputType, readerType, inputFilePath, charset);
+                    continue;
+                }
                 {
```
jdiazgon: Can you check whether this bracket is needed? I think there are too many of them

Mghcapgemini: It is indeed not needed. However it war originally there, maybe for better code comprehension. 
Should I remove it ?

jdiazgon: In my opinion it should be removed, it made it harder for me to understand it. If you find it more readable like that, then you can leave it :)

Mghcapgemini: Okay then, they just got removed :)

cobigen-eclipse/cobigen-eclipse/.classpath
```diff
@@ -20,7 +20,7 @@
 	<classpathentry exported="true" kind="lib" path="lib/mmm-util-core-7.4.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/slf4j-api-1.7.21.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/javaplugin-model-2.0.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/core-5.2.0.jar"/>
-	<classpathentry exported="true" kind="lib" path="lib/core-api-5.2.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-5.3.0-SNAPSHOT.jar" sourcepath="/core"/>
+	<classpathentry exported="true" kind="lib" path="lib/core-api-5.3.0-SNAPSHOT.jar" sourcepath="/core-api"/>
```
jdiazgon: Is this `sourcepath="/core-api` needed? I would remove it if not.

Mghcapgemini: It does. For example the [FileInputConverter](https://github.com/devonfw/tools-cobigen/blob/master/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/generic/FileInputConverter.java) imports some exceptions from the core-api.

jdiazgon: Okay, got it. Thanks :)

maybeec: sourcePath does not have anything to do with imports. It's just an eclipse marker for source lookup in eclipse. So if you navigate on the classpath to a specific class of a dependency, you can specify that your sources are located in a workspace project like /core.
As we are using the same setup for our workspaces, it is anyhow save to be commited. Might make life easier for others in case someone debuggs over library boarders.

PR #922, opened 2019-07-02
Updating external process to latest version

jdiazgon:
Updating external process to latest version. Adding test to check if it is really using the latest ts-merger.

@devonfw/cobigen


jdiazgon:
Our Jenkins is completely broken, so I decided to use a remote Linux machine in order to see if tests are passing, the result is successful:

image

PR #932, opened 2019-07-11
Throw exception when needed

jdiazgon:
The ts-merger lacked a proper exception handling. This PR fixes that problem.

@devonfw/cobigen


jdiazgon:
Seems that production line is broken. I have built this on a Linux machine and all tests are passing, therefore I will merge it.

PR #951, opened 2019-07-17
Jenkins disabling concurrent builds

jdiazgon:
In our current production line, for each new commit pushed to a branch, a new Jenkins build is executed. This is problematic because CobiGen is blocking all the executors of our production line.

Adding property `disableConcurrentBuilds()`, only when the last commit build finishes, the next one will be triggered. 

For instance, I push to `master` commit A, Jenkins triggers a new build. Then I push 3 new commits B, C and D again to `master`. When Jenkins build A finishes, a new build with commits B, C and D will be triggered.

In my opinion, this makes more sense as I only want to know if the most up to date code passes the tests. Also it is worth to mention that it does not affect other branches builds. If I push to `dev_javaplugin` a new build will be triggered.

This has been requested by @dario-rodriguez 

@devonfw/cobigen

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ properties([
     string(name: 'TRIGGER_REPO', defaultValue: '', description: 'The URI of the commit that triggered the calling job')
    ]),
    [$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10']],
-   [disableConcurrentBuilds()]
+   disableConcurrentBuilds()
 ])
 node {
     //lock(resource: "pipeline_${env.NODE_NAME}_${env.JOB_NAME}", inversePrecedence: false) {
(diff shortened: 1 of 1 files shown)


maybeec:
> For instance, I push to master commit A, Jenkins triggers a new build. Then I push 3 new commits B, C and D again to master. When Jenkins build A finishes, a new build with commits B, C and D will be triggered.

This is already the case. Jenkins is triggering a new job per push and not per commit as you wrote.
Anyhow, the only difference would be, that the second push will wait building until the first one is finished, which leads to queuing of builds and execution times of hours.
Btw. honestly spoken, I most often see the executors idle (also now).

There is another issue, which might be more interesting to be discussed. Currently, we are triggering a maven and a eclipse test job after each plugin build as a new plugin version might fail in an integration test within eclipse or maven. At least these two might be worth to be fired, but just queued with a maximum of 1 waiting job. This would prevent from multiple dev_eclipseplugin jobs being queued and processed after another or even in parallel.

>In my opinion, this makes more sense as I only want to know if the most up to date code passes the tests. Also it is worth to mention that it does not affect other branches builds. If I push to dev_javaplugin a new build will be triggered.

So as mentioned, this is acutally not true. The eclipse tests and maven tests are run against the latest snapshot build and thus might fail on plugin changes by intention.

So I basically disaggree to this solution as it does not solve the problem. Is there really a problem of cobigen blocking others?

PR #963, opened 2019-07-23
Adding Automated Unit Test Case

### LINKED ISSUE #947: Add more test cases for CLI
We need to add more automated tests on the CLI so that we don't have to manually test anything. Test cases to add:

- We have an increments and templates search functionality in the CLI. For instance, if I execute `cg g foo\bar\webServices.yml --increment TO` CobiGen is going to print all the increments related to Transfer Objects like TO's, CTO's, CRUD UC Logic (CTO's)... We need to add a JUnit test for both searching increments and templates.

- In the CLI it is possible to use multiple input files of different types. For example, you can execute `cg g foo\employeeEntity.java,foo\webServices.yml`. This performs an increment intersection so that only the increments that are available for both input files will be printed. This needs to be tested also with JUnit.

sanjaykumarcg:
Fixes #947.

Add Unit test case for multiple type input and create method for check generated file is deleted or not .

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/src/test/java/com/devonfw/cobigen/cli/commandtests/GenerateCommandTest.java
+++ b/cobigen-cli/src/test/java/com/devonfw/cobigen/cli/commandtests/GenerateCommandTest.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.assertTrue;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.ArrayList;
 
 import org.apache.commons.io.FileUtils;
 import org.junit.Test;
@@ -18,6 +19,8 @@ public class GenerateCommandTest {
 
     /** Test resources root path */
     private static String testFileRootPath = "src/test/resources/testdata/";
+    /** Declare ArrayList variable for adding generate increment*/
+    ArrayList<File> geneatedList = new ArrayList<File>();
 
     /**
      * Input Java entity used in the tests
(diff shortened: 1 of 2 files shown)


cobigen-cli/src/test/java/com/devonfw/cobigen/cli/commandtests/GenerateCommandTest.java
```diff
@@ -103,24 +106,36 @@ public void generateFromOpenApiTest() throws IOException {
...
+
+		for (File generatedFile : generateFiles) {
+			assertTrue(generatedFile.exists());
+			try {
+				FileUtils.deleteDirectory(generatedFile);
+			} catch (IOException e) {
+				continue;
+			}
+		}
+		return deletedFiles;
```
jdiazgon: This return variable is never used. Please change the method type to `void` and change the name to 'deleteGeneratedFiles'

cobigen-cli/src/test/java/com/devonfw/cobigen/cli/commandtests/GenerateCommandTest.java
```diff
@@ -45,10 +48,8 @@ public void generateFromEntityTest() throws IOException {
 
         File generatedFiles = baseProject.toPath()
             .resolve("src/main/java/com/maven/project/sampledatamanagement/dataaccess/api/repo").toFile();
-
-        assertTrue(generatedFiles.exists());
-        // If you want to remove the generated files
-        FileUtils.deleteDirectory(generatedFiles);
+        geneatedList.add(generatedFiles);
+        geneatedList.clear();
```
jdiazgon: I think here you forgot to call the `deleteGeneratedFiles` method

Issue #956, opened 2019-07-22
--version should print all CobiGen plugins versions


jdiazgon:
### Expected behavior

As a CobiGen CLI user, I want to see which plug-ins versions are used on my CLI so that I can use this information for finding problems.

When running `cg --version`, the CLI should read the artificial pom that is extracted next to the _cobigen.jar_ and print all the plug-ins versions like:

```
$ > cg --version
> CLI 1.0.0
> Java plug-in 2.2.0
> OpenAPI plug-in 2.3.0
> HTML plug-in 2.0.0
...
```

**Note:** The artificial pom is always located next to the _cobigen.jar_. That file defines all the CobiGen plug-ins being used.

![image](https://user-images.githubusercontent.com/31688036/61620190-a836f780-ac70-11e9-85ec-f29e98f9c28f.png)

### Actual behavior
When running `cg --version`, only the version of the CLI is printed.

PR #972, opened 2019-08-22
classloader update

sanjaykumarcg:
Issue : class loader

i done the changes in pom.xml  replace SNAPSHOT with beta2 .
i have maked changes in ParsingUtils.java   and parser.parseQualifiedName(reader, file) replace with parser.parseType(reader, file) and return with getQualifiedName() . 

its returns only simple entity name  like EmployeeEntity ,so its not able to load class ;

### DIFF
--- a/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
+++ b/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
@@ -8,6 +8,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URL;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -16,12 +17,15 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.Properties;
 
+import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.xml.pull.XmlPullParserException;
 
+import com.devonfw.cobigen.cli.constants.MavenConstants;
 import com.devonfw.cobigen.cli.constants.MessagesConstants;
+import com.devonfw.cobigen.cli.utils.CobiGenUtils;
 
 import net.sf.mmm.util.io.api.IoMode;
 import net.sf.mmm.util.io.api.RuntimeIoException;
@@ -46,30 +50,58 @@ public class CobiGenCommand implements Runnable {
 	}
 
 	/**
-	 * This class implement getVersion() and this method return the version of plug-in
+	 * This class implement getVersion() and this method return the version of
+	 * plug-in
 	 */
 	static class PropertiesVersionProvider implements IVersionProvider {
 		@Override
 		public String[] getVersion() throws Exception {
-			String versionProvider[] = new String[50];
+			List<String> versionProvider = new ArrayList<String>();
 			MavenXpp3Reader reader = new MavenXpp3Reader();
-			Model model;
-			if ((new File("pom.xml")).exists())
-				model = reader.read(new FileReader("pom.xml"));
-			else
-				model = reader.read(
-						new InputStreamReader(CobiGenCommand.class.getResourceAsStream("/src/main/resources/pom.xml")));
+			Model model = null;
+			File locationCLI = new File(CobiGenUtils.class.getProtectionDomain().getCodeSource().getLocation().toURI());
+			Path rootCLIPath = locationCLI.getParentFile().toPath();
+
+			File pomFile = extractArtificialPom(rootCLIPath);
+
+			if (pomFile.exists()) {
+				model = reader.read(new FileReader(pomFile));
+			}
+
 			List<Model> versionList = new ArrayList<Model>();
 			versionList.add(model);
 
-			for (int j = 0; j < versionList.get(0).getDependencies().size(); j++) {
-				versionProvider[j] = " name:= " + model.getDependencies().get(j).getArtifactId() + " version=  "
-						+ model.getDependencies().get(j).getVersion();
-			}
+			List<Dependency> modelDependencies = model.getDependencies();
 
-			return versionProvider;
+			for (int i = 0; i < modelDependencies.size(); i++) {
+
+				versionProvider.add(" name:= " + modelDependencies.get(i).getArtifactId() + " version=  "
+						+ modelDependencies.get(i).getVersion());
+
+			}
+			return versionProvider.toArray(new String[versionProvider.size()]);
 		}
 
+	}
+
+	/**
+	 * Extracts an artificial POM which defines all the CobiGen plug-ins that are
+	 * needed
+	 * 
+	 * @param rootCLIPath path where the artificial POM will be extracted to
+	 * @return the extracted POM file
+	 */
+	private static File extractArtificialPom(Path rootCLIPath) {
+		File pomFile = rootCLIPath.resolve(MavenConstants.POM).toFile();
+		if (!pomFile.exists()) {
+			try (InputStream resourcesIS = (CobiGenCommand.class.getResourceAsStream("/" + MavenConstants.POM));) {
+				Files.copy(resourcesIS, pomFile.getAbsoluteFile().toPath());
+			} catch (IOException e1) {
+				System.out.println(
+						"Failed to extract CobiGen plugins pom into your computer. Maybe you need to use admin permissions.");
+			}
+		}
+		return pomFile;
 	}
 
 }
--- a/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
+++ b/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
@@ -8,7 +8,7 @@ import picocli.CommandLine.Command;
  * This class defines the main CobiGen command
  */
 @Command(description = MessagesConstants.WELCOME_MESSAGE, name = "cobigen", aliases = { "cg" },
-    mixinStandardHelpOptions = true, version = {"CobiGen CLI 1.0","Java plug-in 2.1.0","OpenAPI plug-in	2.3.0","HTML plug-in 2.0.1"}, subcommands = { GenerateCommand.class })
+    mixinStandardHelpOptions = true, version = "CobiGen CLI 1.0", subcommands = { GenerateCommand.class })
 public class CobiGenCommand implements Runnable {
 
     @Override
--- a/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
+++ b/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
@@ -45,10 +45,13 @@ public class CobiGenCommand implements Runnable {
 		// Nothing to do here, this is the master command
 	}
 
+	/**
+	 * This class implement getVersion() and this method return the version of plug-in
+	 */
 	static class PropertiesVersionProvider implements IVersionProvider {
 		@Override
 		public String[] getVersion() throws Exception {
-
+			String versionProvider[] = new String[50];
 			MavenXpp3Reader reader = new MavenXpp3Reader();
 			Model model;
 			if ((new File("pom.xml")).exists())
@@ -56,17 +59,15 @@ public class CobiGenCommand implements Runnable {
 			else
 				model = reader.read(
 						new InputStreamReader(CobiGenCommand.class.getResourceAsStream("/src/main/resources/pom.xml")));
-			List<Model> versionname = new ArrayList<Model>();
-			versionname.add(model);
-			for (int i = 0; i < versionname.size(); i++)
-				for (int j = 0; j < versionname.get(i).getDependencies().size(); j++) {
-					System.out.println("name:= "+versionname.get(i).getName()+"  version"+versionname.get(i).getVersion());
-					System.out.println("name:= " + versionname.get(i).getDependencies().get(j).getArtifactId()
-							+ "  version= " + versionname.get(i).getDependencies().get(j).getVersion());
-				}
-			String s[] = new String[10];
+			List<Model> versionList = new ArrayList<Model>();
+			versionList.add(model);
 
-			return s;
+			for (int j = 0; j < versionList.get(0).getDependencies().size(); j++) {
+				versionProvider[j] = " name:= " + model.getDependencies().get(j).getArtifactId() + " version=  "
+						+ model.getDependencies().get(j).getVersion();
+			}
+
+			return versionProvider;
 		}
 
 	}
--- a/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
+++ b/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
@@ -1,20 +1,74 @@
 package com.devonfw.cobigen.cli.commands;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.URL;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Properties;
+
+import org.apache.maven.model.Model;
+import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.shared.utils.xml.pull.XmlPullParserException;
+
 import com.devonfw.cobigen.cli.constants.MessagesConstants;
 
+import net.sf.mmm.util.io.api.IoMode;
+import net.sf.mmm.util.io.api.RuntimeIoException;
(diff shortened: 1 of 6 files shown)


cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -385,7 +384,7 @@ private void setOutputRootPath(File inputProject) {
                         : ((TemplateTo) matching.get(index)).getId();
                     logger.info("(" + selectedArtifactNumber + ") " + artifactDescription);
                 } catch (IndexOutOfBoundsException e) {
-                    logger.error("The " + artifactType + " number you have specified is out of bounds!");
+                    logger.error("The " + artifactType + " number you have specified is out of bounds ,So please provide the number(s) of increments that you want to generate separated by comma.!");
```
jdiazgon:
```Java
So please provide the number(s) of increments that you want to generate separated by comma
```
That message doesn't make sense to me. This exception is thrown when the user selects an increment number that doesn't exist (for instance `459`, we only have 15 increments at most). Therefore the message you have added is not needed

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -421,10 +420,14 @@ private void setOutputRootPath(File inputProject) {
                     logger.info("Exact match found: " + artifactDescription + ".");
                     userSelection.add(possibleArtifacts.get(0));
                     return userSelection;
+                }else if(possibleArtifacts.size()<1) {
+                	logger. info("No increment with that name has been found, Please provide correct increment name and try again ! Thank you");
```
jdiazgon: Great, this is what I wanted :)

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -326,13 +326,12 @@ public void generate(File inputFile, File inputProject, List<? extends Generable
      * @param inputProject
      *            project where the code will be generated to
      */
-    private void setOutputRootPath(File inputProject) {
-        logger.info("As you did not specify where the code will be generated, we will use the project of your current"
-            + " input file.");
-        logger.debug("Generating to: " + inputProject.getAbsolutePath());
```
jdiazgon: Please don't remove these two messages, they are important to have

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -421,10 +420,14 @@ private void setOutputRootPath(File inputProject) {
...
+                	logger. info("No increment with that name has been found, Please provide correct increment name and try again ! Thank you");
+                       
+                        System.exit(1);
                 }
-
-                logger.info("Please enter the number(s) of " + artifactType
-                    + "(s) that you want to generate separated by comma.");
+                
+                /*logger.info("Please enter the number(s) of " + artifactType
+                    + "(s) that you want to generate separated by comma.");*/
```
jdiazgon: Can you explain me why have you removed this message?

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -364,25 +365,25 @@ public File getTemplatesJar(boolean isSource) {
...
      *            list containing elements to be retained in this list
-     * @param listToIntersect
+     * @param list2
      *            second list to be used for the intersection
      * @return <tt>resultant list</tt> containing increments that are in both lists
      */
-    public static List<IncrementTo> retainAllIncrements(List<IncrementTo> currentList,
-        List<IncrementTo> listToIntersect) {
+    public static List<IncrementTo> retainAllIncrements(List<IncrementTo> list,
+        List<IncrementTo> list2) {
```
jdiazgon: Why have you changed the names of the variables? They had good names previously

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -57,19 +57,49 @@
...
 
-        JavaContext context = JavaSourceProviderUsingMaven.createFromLocalMavenProject(inputProject);
-
+        JavaContext context = JavaSourceProviderUsingMaven.createFromLocalMavenProject(inputProject,true);
+        String fqn = ParsingUtils.getFQN(inputFile);
         String qualifiedName = ParsingUtils.getQualifiedName(inputFile, context);
 
         try {
-            context.getClassLoader().loadClass(qualifiedName);
+           // context.getClassLoader().loadClass(qualifiedName);
```
jdiazgon: Please remove this comment

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -57,19 +57,49 @@
...
+     * @return
+     */
+    private static String getFQN(File inputFile) {
+        String simpleName = inputFile.getName().replaceAll("\\.(?i)java", "");
+        String packageName = getPackageName(inputFile.getParentFile(), "");
+
+         return packageName + "." + simpleName;
+    }
+
+     private static String getPackageName(File folder, String packageName) {
```
jdiazgon: Can you add Javadoc? This method is traversing parent folders until it reaches `java` folder in order to get the FQN

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
sanjaykumarcg: Actually this message is comes under If String representation is given,so dont need this message here .

jdiazgon: You are right that it comes under String representation, but please don't remove this message because in this part of the code (line 415):

```Java
if (possibleArtifacts.size() > 1) {
    printFoundArtifacts(possibleArtifacts, isIncrements, artifactType);
} 
```
We are printing all the found artifacts by number, like this:

```
(0) All
(1) TO's
(2) CRUD Rest services
...
```
Therefore that message makes sense, because the user needs to use numbers to select the option.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -384,7 +386,7 @@ private void setOutputRootPath(File inputProject) {
                         : ((TemplateTo) matching.get(index)).getId();
                     logger.info("(" + selectedArtifactNumber + ") " + artifactDescription);
                 } catch (IndexOutOfBoundsException e) {
-                    logger.error("The " + artifactType + " number you have specified is out of bounds ,So please provide the number(s) of increments that you want to generate separated by comma.!");
+                    logger.error("The " + artifactType + " number you have specified is out of bound !");
```
jdiazgon: It's `out of bounds` you need the final `s` ;)

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -185,7 +185,7 @@ public CobiGen initializeCobiGen() {
         CobiGen cg = null;
         if (templatesJar != null) {
             try {
-                registerPlugins();
+            	registerPlugins();
```
jdiazgon: Here you have added some extra tabs that we don't need, please remove those whitespaces

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -59,30 +59,34 @@ public static JavaContext getJavaContext(File inputFile, File inputProject) {
...
     }
-
-     private static String getPackageName(File folder, String packageName) {
+/**
+ *  This method traverse the folder in reverse order from child to parent
+ *  @param folder parent input file
+ *  @param package name 
+ *  @return package name
+ * */
+     @SuppressWarnings("javadoc")
```
jdiazgon: The javadoc is perfect, but I don't understand why you have added `@SuppressWarnings("javadoc")`. Please remove it

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -8,7 +8,7 @@
  * This class defines the main CobiGen command
  */
 @Command(description = MessagesConstants.WELCOME_MESSAGE, name = "cobigen", aliases = { "cg" },
-    mixinStandardHelpOptions = true, version = "CobiGen CLI 1.0", subcommands = { GenerateCommand.class })
+    mixinStandardHelpOptions = true, version = {"CobiGen CLI 1.0","Java plug-in 2.1.0","OpenAPI plug-in	2.3.0","HTML plug-in 2.0.1"}, subcommands = { GenerateCommand.class })
```
jdiazgon: This is related to issue #956, right? Actually that is not enough, please read again the issue.

What we need to do is read the artificial pom, in order to parse the different versions of the plug-ins and add them there. If you have any question regarding this, contact me.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -178,29 +179,29 @@ public FileVisitResult visitFileFailed(Path file, IOException exc) throws IOExce
      * Registers CobiGen plug-ins and instantiates CobiGen
      * @return object of CobiGen
      */
-    public CobiGen initializeCobiGen() {
-        getTemplates();
+	public CobiGen initializeCobiGen() {
```
jdiazgon: Hmm these tab characters doesn't look good. Can you revert these changes?

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -249,7 +251,7 @@ public void registerPlugins() {
             logger.error("Not able to convert current location of the CLI to URI. Most probably this is a bug", e);
         }
 
-    }
+    };
```
jdiazgon: This `;` is not actually needed.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -364,9 +366,9 @@ public File getTemplatesJar(boolean isSource) {
      * specified collection (optional operation). In other words, the resultant list removes from this list
      * all of its elements that are not contained in the specified collection.
      *
-     * @param currentList
+     * @param list
      *            list containing elements to be retained in this list
-     * @param listToIntersect
+     * @param list2
```
jdiazgon: Please change the JavaDoc back to `currentList` and `listToIntersect`

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -48,19 +57,53 @@
...
+        String packageName = getPackageName(inputFile.getParentFile(), "");
+
+         return packageName + "." + simpleName;
+    }
+/**
+ *  This method traverse the folder in reverse order from child to parent
+ *  @param folder parent input file
+ *  @param packageName 
+ *  @return package name
+ * */
```
jdiazgon: Please format this correctly, the JavaDoc needs to be aligned with the method declaration.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -1,20 +1,74 @@
...
+			MavenXpp3Reader reader = new MavenXpp3Reader();
+			Model model;
+			if ((new File("pom.xml")).exists())
+				model = reader.read(new FileReader("pom.xml"));
+			else
+				model = reader.read(
+						new InputStreamReader(CobiGenCommand.class.getResourceAsStream("/src/main/resources/pom.xml")));
+			List<Model> versionname = new ArrayList<Model>();
+			versionname.add(model);
+			for (int i = 0; i < versionname.size(); i++)
```
jdiazgon: I think you don't need this `for` loop.

So you have created a List called `versionname` where you add the model: `versionname.add(model)`

Why don't you directly use `model` itself in order to get the dependencies?

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -1,20 +1,74 @@
...
+				for (int j = 0; j < versionname.get(i).getDependencies().size(); j++) {
+					System.out.println("name:= "+versionname.get(i).getName()+"  version"+versionname.get(i).getVersion());
+					System.out.println("name:= " + versionname.get(i).getDependencies().get(j).getArtifactId()
+							+ "  version= " + versionname.get(i).getDependencies().get(j).getVersion());
+				}
+			String s[] = new String[10];
```
jdiazgon: It seems that you need to add to that array all the different artifacts and versions, like:

```Java
s[0] = ""name:= " + versionname.get(i).getDependencies().get(j).getArtifactId()
							+ "  version= " + versionname.get(i).getDependencies().get(j).getVersion())
```

Therefore, you don't need those `System.out.println` as Picocli will use that array in order to print the versions

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -1,20 +1,74 @@
...
+		@Override
+		public String[] getVersion() throws Exception {
+
+			MavenXpp3Reader reader = new MavenXpp3Reader();
+			Model model;
+			if ((new File("pom.xml")).exists())
+				model = reader.read(new FileReader("pom.xml"));
+			else
+				model = reader.read(
+						new InputStreamReader(CobiGenCommand.class.getResourceAsStream("/src/main/resources/pom.xml")));
```
jdiazgon: Actually, we need to read the pom.xml that is present on the cobigen-cli folder. On the first execution of the CLI, we extract that pom.xml next to the `cobigen.jar` file. You should read that pom, because in the future, we will be able to change versions directly to that pom.

If you take a look [here](https://github.com/devonfw/tools-cobigen/blob/master/cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java#L230), you will see how to get the real path of the pom.xml.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -1,20 +1,74 @@
...
+		// Nothing to do here, this is the master command
+	}
+
+	static class PropertiesVersionProvider implements IVersionProvider {
+		@Override
+		public String[] getVersion() throws Exception {
+
+			MavenXpp3Reader reader = new MavenXpp3Reader();
+			Model model;
+			if ((new File("pom.xml")).exists())
```
jdiazgon: Is this always returning `false`? If so, please look on the comment below.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java
```diff
@@ -375,20 +376,21 @@ public File getTemplatesJar(boolean isSource) {
...
+				String intersectIncrementDesc = ((IncrementTo) intersectIncrement).getDescription().trim()
+						.toLowerCase();
+
+				if (currentIncrementDesc.equals(intersectIncrementDesc)) {
+					resultantList.add((IncrementTo) currentIncrement);
+					break;
+				}
+			}
+		}
+		return resultantList;
```
jdiazgon: Can you revert these changes? I think they are not needed.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -46,30 +50,58 @@ public void run() {
...
+			List<String> versionProvider = new ArrayList<String>();
 			MavenXpp3Reader reader = new MavenXpp3Reader();
-			Model model;
-			if ((new File("pom.xml")).exists())
-				model = reader.read(new FileReader("pom.xml"));
-			else
-				model = reader.read(
-						new InputStreamReader(CobiGenCommand.class.getResourceAsStream("/src/main/resources/pom.xml")));
+			Model model = null;
+			File locationCLI = new File(CobiGenUtils.class.getProtectionDomain().getCodeSource().getLocation().toURI());
```
jdiazgon: Instead of `CobiGenUtils` can you use `CobiGenCommand`? It is the class that we currently are :)

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/CobiGenCommand.java
```diff
@@ -46,30 +50,58 @@ public void run() {
...
 	}
 
+	/**
+	 * Extracts an artificial POM which defines all the CobiGen plug-ins that are
+	 * needed
+	 * 
+	 * @param rootCLIPath path where the artificial POM will be extracted to
+	 * @return the extracted POM file
+	 */
+	private static File extractArtificialPom(Path rootCLIPath) {
```
jdiazgon: Here you are duplicating this method from CobiGenUtils. Please call directly the method from there, make it `public` if needed

PR #977, opened 2019-09-02
Licensing OSS tsplugin

jdiazgon:
Addressing license issues for our plug-ins:

- Removed old `LICENSE.CobiGen`
- Added new license which is included only on the P2 bundle.
- On Maven, we will use Apache license.

@devonfw/cobigen


jdiazgon:
1. When running `mvn clean install -DskipTests` the created jar on `target\tsplugin-2.4.0-SNAPSHOT.jar` should contain the Apache license (it is for Maven Central). Please add the Apache License if it is not found under `src/main/resources/META-INF/LICENSE.txt`
2. When running `mvn install -U bundle:bundle -Pp2-bundle p2:site -Dmaven.test.skip=true` the created jar on `target\repository\plugins\com.devonfw.cobigen.tsplugin_2.4.0.jar` should contain a large file License.html with a lot of licenses (it is a P2 bundle for Eclipse plugins)

PR #976, opened 2019-08-29
External TS input reader

Mghcapgemini:
Adresses/Fixes #914.

Implements

* A Server-client model based typescript input reader with,
  * typescript matcher and trigger interpreter. 
  * Tests and test files.

@devonfw/cobigen

PR #978, opened 2019-09-06
devon4node templates: generate Angular client

jdiazgon:
These templates generate an Angular client from [Typeorm ](https://typeorm.io/#/) entities which are used on devon4node. This should be merged after #976 PR.

The Angular client has been implemented to be compatible with the [devon4node](https://github.com/devonfw/devon4node/pull/15) back-end.

@devonfw/cobigen


jdiazgon:
I have manually tested them and they are working fine. Therefore I'm merging this PR.

PR #994, opened 2019-09-13
Bug fix on server - tsplugin

jdiazgon:
We found an issue on dev_core: the problem was when starting the external process, it threw a NPE because in my last PR I added some parameters on the URL that were null. Our tsplugin tests didn't show that error because they initialize the server prior to the tests. This pull request fixes those tests.

@devonfw/cobigen


jdiazgon:
Again, on my linux machine I'm getting no errors on the tests:

image

PR #998, opened 2019-09-17
Changing property on pom.xml

jdiazgon:
Property `${artifactId}` was not working when creating the p2 repository jar. It didn't generate `tsplugin` but `${artifactId}`.

@devonfw/cobigen


jdiazgon:
I just don't get why are we getting those test failures. Our production line instance seems to do weird things.

Issue #914, opened 2019-06-18
CobiGen Refactoring: TypeScript input reader


jdiazgon:
### Description
Related to the CobiGen refactoring, in which we are implementing inter process communication (IPC), we need to implement a TypeScript input reader that is able to parse TS code into a valid model for our Freemarker templates engine.

This needs to be implemented on both the [server ](https://github.com/jdiazgon/NodeServer/tree/dev) side and the CobiGen side.

jdiazgon:
I already started to work on this, I pushed a commit including some tests and the initial version of the input reader. Please cherry-pick [the commit](https://github.com/jdiazgon/tools-cobigen/commit/9fa95a790760d2ad3e80da29428015fbe903e64a).

Also note that the transfer object of the input reader should be changed. The property `String path` from [here](https://github.com/jdiazgon/tools-cobigen/commit/9fa95a790760d2ad3e80da29428015fbe903e64a), should be changed to `String contents`. Per customer request, we should not send the path of the file, but the complete file content as a string. Just as we do currently in the ts-merger.

In the server side, I also started to implement some things. Please take a look to [this commit](https://github.com/jdiazgon/NodeServer/commit/39b9ebd3927487737d0e30f035e0a61543dbc827). I found a very interesting library called [typescript-parser](https://github.com/buehler/node-typescript-parser), this tool parses TS code and returns a human readable model. We could use this library in order to create our models and return them to our CobiGen input reader. CobiGen will then pass this model to FreeMarker and generate templates.

Mghcapgemini:
As the [typescript-parser](https://github.com/buehler/node-typescript-parser) library miss some features needed for [type-orm](https://typeorm.io/) entity parsing, such as the extraction of decorators and variable values, which are however already provided by the [ts-merger](https://github.com/devonfw/ts-merger), I decided to base the typescript input reading on the latter library and adapt the extracted AST to match the given specifications and be human readable.

PR #1000, opened 2019-09-17
Generic plug-in for refactoring

jdiazgon:
### Description
Related to the CobiGen refactoring, in which we are implementing inter process communication (IPC), we have already implemented a TypeScript input reader (see #914) and merger on an external server.

One of the requirements of this refactoring is to ease the creation of new CobiGen plug-ins. This PR contains a template plugin (`todoplugin`) which the developer will just have to rename it to the correct plugin name and modify some properties on the pom.xml file.

I thought that it would be much easier to maintain it if we pushed it to our main repo. Therefore, it will allow us to create pull requests also for this "template" plug-in. I think it is okay to add it to `master`, but maybe @maybeec prefers it to be on a new `dev_todoplugin` branch. What do you think?

@devonfw/cobigen


jdiazgon:
Plugin #1000, cool number for a feature like this one 😄

jdiazgon:
After a discussion with Malte, we have decided to try a new feature of GitHub called templates repository. I'm going to close this PR meanwhile.

PR #1004, opened 2019-09-24
[TEST] Fixing Jenkins pipeline

jdiazgon:
On our Jenkins pipeline, `tsplugin` tests are always failing even though in my local they work fine. With this PR, I'm trying to fix those issues.

@devonfw/cobigen


jdiazgon:
Finally the pipeline is now working. It seems that what we had to do was removing old external process files.

PR #1046, opened 2019-10-31
Args with quote support

### LINKED ISSUE #1045: cobigen-cli shell wrappers fail to handle args with quote
### Expected behavior

As a user, I want to use cobigen-cli in a way so that even arguments containing quotes are handled correctly.

### Actual behavior

cobigen / cg separate arguments even with quoted spaces and such fail to identify files with space in path or filename.

```
$ ./cg g "openapi file.xml"
[ERROR] Your <inputFile> 'openapi' has not been found
```

### Comments/Hints:

Use `"$@"` instead of `$*`. This forwards the correct quoting of arguments to other called programs. Hint: Use https://www.shellcheck.net/ to detect such issues.

Also you might replace the current check, if any argument is added, with a more often used code - even when the current check does not run into errors when quoted args are used.

Use either
```
if [ $# -eq 0 ] . . .
if [ -z "$1" ] . . .
```

Compare with #1044

Mghcapgemini:
Adresses/Fixes #1045.

Implements

* Support of args with double " or single ' quotes. 

@devonfw/cobigen


cobigen-cli/cli/src/test/resources/testdata/openAPI file.yml
```diff
@@ -0,0 +1,111 @@
+openapi: 3.0.0
```
jdiazgon: Why don't you use the yml file we already have? Please see [here](https://github.com/devonfw/tools-cobigen/blob/master/cobigen-cli/cli/src/test/resources/testdata/openAPI.yml)

Mghcapgemini: done

Issue #1047, opened 2019-10-31
cobigen CLI can only be called within devonfw context


markusschuh:
### Expected behavior

As a devonfw user I want to use cobigen-cli even outside of devonfw shells - if possible - so that I can get the most win out of it.

### Actual behavior

The current shell wrappers only work, if an environment variable `SOFTWARE_PATH` exists and is set to the path of devonfw-ide generated environment.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

Run one of `cg / cobigen / cg.bat / cobigen.bat`, when `SOFTWARE_PATH` is not defined inside the environment and you'll run into error messages.

### Comments/Hints:

The wrappers are to be saved inside the same directory as the `*.jar` file, that is run with help of `java`, aren't them?
So the "directory of the script file" can be used instead of `$SOFTWARE_PATH/cobigen-cli` resp.  `%SOFTWARE_PATH%/cobigen-cli`. At the same time, the path should be quoted with double quotes in order to support spaces inside the installation path.

I suggest to change inside the both `*.bat` files
```
java -javaagent:%SOFTWARE_PATH%\cobigen-cli\class-loader-agent.jar -jar %SOFTWARE_PATH%\cobigen-cli\cobigen.jar ...
```

into 

```
java -javaagent:"%~dp0\class-loader-agent.jar" -jar "%~dp0\cobigen.jar" ...
```

and preceed this call within the both `bash` wrappers with a variable definition like

```
my_script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
and then use this for the later java calls like
```
java -javaagent:"$my_script_dir"/class-loader-agent.jar -jar "$my_script_dir"/cobigen.jar ...
```

I suggest you'll check your bash files with help of some lint tools, to identify those issues. E.g. for  OSS scripts - and especially short ones, such as for the cobigen wrappes - you can even use the online check at https://www.shellcheck.net/.  But shellcheck may also be used offline - or during travis checks, as in devonfw-ide.

jdiazgon:
>The wrappers are to be saved inside the same directory as the *.jar file, that is run with help of java, aren't them?

Exactly, our bash files should be stored on the same directory as the *.jar. I agree with your point.

markusschuh:
Fixed by #1051. No more dependency on "SOFTWARE_PATH" environment variable. Thx.

PR #1051, opened 2019-11-05
Cli wrappers enhancement

### LINKED ISSUE #1044: cobigen-cli CMD wrappers fail to handle args with quote
### Expected behavior

As a user, I want to use cobigen-cli in a way so that even arguments containing quotes are handled correctly.

### Actual behavior

cobigen.bat / cg.bat run into an error, when quotes are used in command line call.

```
...\cobigen.bat g "openapi file.xml"
file.xml""=="" was unexpected at this time.
```

### Comments/Hints:

In order to check in windows CMD, if **any** argument was provided, check only for the first - and avoid to use quotes for that check - or if, then ensure, the quotes in arg are removed.
Instead use **one** of the following examples:

```
IF [%1]==[] . . .
IF #%1==# . . .
IF "%~1"=="" . . .
```

The first is probably the best one. It is listed at https://ss64.com/nt/if.html which is a good reference for developing CMD scripts in my experience.

Mghcapgemini:
Adresses/Fixes #1044, #1045, #1047 

Implements

* Bugfixes of the cli's CMD and shell wrappers
* Support of cli calls outside of devon environment

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/cli/src/main/resources/batchFiles/cg
+++ b/cobigen-cli/cli/src/main/resources/batchFiles/cg
@@ -1,7 +1,7 @@
 #!/bin/bash
-
-if [ -z "$*" ]; then
-    java -jar $SOFTWARE_PATH/cobigen-cli/cobigen.jar -help
+MY_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+if [ -z "$1" ]; then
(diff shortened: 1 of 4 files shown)


jdiazgon:
Hi @markusschuh,

from my point of view this already addresses the issues you mentioned. But please check it just in case-

PR #1052, opened 2019-11-05
Fixing CLI tests JDK 9+

jdiazgon:
Fixing some CLI tests for JDK 9+.

@devonfw/cobigen


cobigen-cli/cli/pom.xml
```diff
@@ -163,20 +163,16 @@
             </arguments>
          </configuration>
       </plugin>
-      <plugin>
-         <artifactId>maven-compiler-plugin</artifactId>
-         <configuration>
-            <source>1.8</source>
-            <target>1.8</target>
-         </configuration>
-      </plugin>
```
maybeec: changing this, would break backward compatibility for java 8 most probably

cobigen-cli/class-loader-agent/pom.xml
```diff
@@ -10,12 +10,12 @@
     <version>dev-SNAPSHOT</version>
   </parent>
   
-  
   <build>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <version>3.1.2</version>
```
maybeec: is this a newer version? if so maybe you can upgrade it on the master pom to not let every plugin overwrite it again on it's own

cobigen-cli/cli/pom.xml
maybeec: or check wether 

```xml
<properties>
    <maven.compiler.source>1.8</maven.compiler.source>
    <maven.compiler.target>1.8</maven.compiler.target>
  </properties>
```

is set correctly in the properties, which would make this code redundant. Then it's ok to remove it.

jdiazgon: I have printed all the properties to a console and it seems it is not set. I'm going to revert this change

PR #1054, opened 2019-11-07
checked-in for formatting code generation

### LINKED ISSUE #1049: CobiGen CLI formatting issues
As a CobiGen CLI user, I want the generated Java code to be correctly formatted so that I don't have to clean the code after generating.

### Description

When generating Java code using the CobiGen CLI, there are a lot of formatting issues as we don't have an automatic Java code formatter. For our CobiGen-Eclipse plug-in, we automatically run the Eclipse formatter, that's why we don't have that issue there.

My proposal is to use a library to automatically format Java code. Please check [this link](https://stackoverflow.com/a/47634476).

The idea will be:

1. User generates Java code. 
2. After the generation is successful (the generation report is successful), we read from the generation report which files have been just generated. We run that library to format the Java code.

### Reproduce issue

1. Use any Java entity and generate increments like `CRUD REST Services, TOs, CRUD UC Impl...`
2. You will see a lot of formatting issues.

sanjaykumarcg:
Adresses/Fixes #1049 .

I have done changes in GenerateCommand.java,ValidationUtils.java,pom.xml and call google api formatSource() method.

@devonfw/cobigen


cobigen-cli/pom.xml
```diff
@@ -73,6 +73,26 @@
...
+<dependency>
+    <groupId>com.google.guava</groupId>
+    <artifactId>guava</artifactId>
+    <version>23.6-jre</version>
+</dependency> 
+<dependency>
+    <groupId>org.apache.httpcomponents</groupId>
+    <artifactId>httpcore</artifactId>
+    <version>4.4.8</version>
+</dependency>
```
jdiazgon: Please format correctly the pom file. The indentation should be the same for every dependency.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -317,13 +323,40 @@ public void generate(File inputFile, File inputProject, List<? extends Generable
...
+                            String unformatterContent = IOUtils.toString(inputStream);
+                            new Formatter().formatSource(unformatterContent);
+                        } catch (IOException e) {
+                            logger.error("Unable to read input file");
+                        }
+                    } catch (FormatterException e) {
+                        logger.error("No elements found");
+                    }
+                }
+
```
jdiazgon: Please create a new method called `formatJavaSources(Set<Path> generatedFiles)` and move it to utilities class `ParsingUtils`

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -317,13 +323,40 @@ public void generate(File inputFile, File inputProject, List<? extends Generable
...
             }
-            ValidationUtils.checkGenerationReport(report);
+            if (isJavaInput && ValidationUtils.checkGenerationReport(report)) {
+                Set<Path> format = report.getGeneratedFiles();
+                Iterator<Path> itr = format.iterator();
+                if (itr.hasNext()) {
+                    try {
+
+                        try {
+                            InputStream inputStream = Files.newInputStream(itr.next());
```
jdiazgon: Remove the `try` from line 330 as it is not needed and implement a closeable `try` like:

`try (InputStream inputStream = Files.newInputStream(itr.next())) { ..... }`

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -317,13 +323,40 @@ public void generate(File inputFile, File inputProject, List<? extends Generable
...
 
+    /**
+     * @param file
+     * @return
+     */
+    private InputStream newInputStream(File file) {
+        // TODO Auto-generated method stub
+        return null;
+    }
+
```
jdiazgon: Please remove this method

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -317,13 +323,40 @@ public void generate(File inputFile, File inputProject, List<? extends Generable
...
+            if (isJavaInput && ValidationUtils.checkGenerationReport(report)) {
+                Set<Path> format = report.getGeneratedFiles();
+                Iterator<Path> itr = format.iterator();
+                if (itr.hasNext()) {
+                    try {
+
+                        try {
+                            InputStream inputStream = Files.newInputStream(itr.next());
+                            String unformatterContent = IOUtils.toString(inputStream);
+                            new Formatter().formatSource(unformatterContent);
```
jdiazgon: You are creating a new `Formatter` for each loop iteration. Please move the declaration before the iteration and reuse the same instance.

Also here you are formatting the code but you are not replacing the old contents of the file. You need to remove the old file contents and replace them with the formatted string

cobigen-cli/pom.xml
```diff
@@ -73,6 +73,26 @@
...
+    <dependency>
+  <groupId>com.google.googlejavaformat</groupId>
+  <artifactId>google-java-format</artifactId>
+  <version>1.7</version>
+</dependency>
+<dependency>
+  <groupId>com.google.googlejavaformat</groupId>
+  <artifactId>google-java-format</artifactId>
+  <version>1.7</version>
+</dependency>
```
jdiazgon: You have duplicated the same dependency, please remove one.

cobigen-cli/pom.xml
```diff
@@ -73,6 +73,26 @@
...
+<dependency>
+    <groupId>com.google.guava</groupId>
+    <artifactId>guava</artifactId>
+    <version>23.6-jre</version>
+</dependency> 
+<dependency>
+    <groupId>org.apache.httpcomponents</groupId>
+    <artifactId>httpcore</artifactId>
+    <version>4.4.8</version>
```
jdiazgon: Are really these dependencies needed? Please check that

maybeec:
does this formatter take anything into account? 
For example standards like .editorconfig? https://editorconfig.org/
Otherwise, CLI will always reformat code which might be differently formatted as the rest of the code.

jdiazgon:
> does this formatter take anything into account?

From their README:

`google-java-format` is a program that reformats Java source code to comply with
[Google Java Style][].

[Google Java Style]: https://google.github.io/styleguide/javaguide.html

Basically the problem is that our templates doesn't always use the correct indentation, or methods are too close, or too far between each other. This Google Java format will help to have a much cleaner code.

It's true that if a developer uses a different java format standard (for instance: 6 white spaces instead of 4 or three blank lines between methods), then the generated code won't be the same. But at least our Google Java formatter will clean the code and follow a Java format standard. 

IMHO, it's better than having the dirty generated code and it means we don't have to worry about our templates using the correct formatting.

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -213,4 +221,30 @@ public static File getProjectRoot(File inputFile) {
...
+     *            List of generation report files
+     * @throws FormatterException
+     */
+    public static void formatJavaSources(Set<Path> generatedFiles) throws FormatterException {
+        Set<Path> format = generatedFiles;
+        Formatter formatter = new Formatter();
+        Iterator<Path> itr = format.iterator();
+        while (itr.hasNext()) {
+            try {
+                InputStream inputStream = Files.newInputStream(itr.next());
```
jdiazgon: please remember the closeable `try (){}`. Also please fix the merge conflicts

cobigen-cli/src/main/java/com/devonfw/cobigen/cli/utils/ParsingUtils.java
```diff
@@ -213,4 +221,30 @@ public static File getProjectRoot(File inputFile) {
...
+                String unformatterContent = IOUtils.toString(inputStream);
+                String formatedCode = formatter.formatSource(unformatterContent);
+                FileOutputStream File = new FileOutputStream(inputStream.toString());
```
jdiazgon: In a call we already saw that this was not correctly saving the file to the right path. Please check a way to replace the old unformatted file with the formatted contents

jdiazgon:
Replaced by #1062

PR #1056, opened 2019-11-08
Remove deprecated M2_HOME environment variable

jdiazgon:
Due to devon-ide requirement on PR [#273](https://github.com/devonfw/ide/pull/273), where they requested not to use `M2_HOME` variable as it is deprecated, I create this PR to fix that issue.

I programmatically set `MAVEN_HOME` as the default variable.

@devonfw/cobigen


jdiazgon:
I have tested it on both Linux and Windows. Working fine.

PR #1058, opened 2019-11-13
Dev eclipseplugin

### LINKED ISSUE #867: Cobigen processbar
### Expected behavior
When i use cobigen to generate code i want to see how long it takes until it is finished and i want the option to cancel the code generation (maybe if it takes too long). 

### Actual behavior
At the moment we have a processbar but it is not working as expected. Wenn i generate some code there is nothing happen and you can't use the cancel button its disabled. Not before the generation is nearly allready finished there appear a few more windows and then the generation is finished. Especially if i generate a lage amount of code, i have to wait a long time and don't know if it is still working or if eclipse suspended completly.

grafik

LarsReinken:
Adresses/Fixes #867.

Implements

* 
* 
* 

@devonfw/cobigen


cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/wizard/generate/control/GenerateSelectionJob.java
```diff
@@ -155,6 +157,8 @@ public void run() {
                     LOG.error("An error occurred during generation:", e);
                 }
             }
+        } catch (CancellationException e) {
+            LOG.error("generation got cancelled", e);
```
maybeec: The cancellationException will never occur up to this point right?

PR #1107, opened 2020-02-18
Updated FreeMarker to 2.3.29

jan-vcapgemini:
Adresses/Fixes #<ISSUENO>.

Implements

*  latest changes from https://freemarker.apache.org/docs/versions_2_3_29.html

@devonfw/cobigen


maybeec:
@jdiazgon, thanks for merging. Please make sure, that you add such things to the milestone in this case. I have seen we missed to maintain a proper changelog in the past which is quite odd. THanks.

Issue #1083, opened 2020-01-08
Failing tests due to JDK 11 version


jdiazgon:
### Description

Due to the migration from JDK 8 to JDK 11 there are some plug-ins which tests are failing. We should check these issues for next release.

### Affected plug-ins:
1. Property plug-in.
2. Text merger.
3. Java plug-in.
4. XML plug-in
5. OpenAPI plug-in

maybeec:
@jdiazgon, I fixed quite a long and implemented travis builds and deployment.
The tests are now running on openjdk8 + openjdk11 all the time. Unfortunately, I see a lot of tests failing especially with openjdk8. So we need to get this fixed as well. Please find the builds here: 
https://travis-ci.com/github/devonfw/cobigen/branches

jdiazgon:
Thanks @maybeec for your great support. Our colleague @sujithmn is working on them.

PR #1132, opened 2020-04-08
Dev jdk11issues

sujith-mn:
Adresses/Fixes #<ISSUENO>.

Implements

Fixed the build failures due to the compilation issues on the below moduels:

- javaplugin
- propertyplugin
- textmerger

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-javaplugin-parent/cobigen-javaplugin/src/test/java/com/devonfw/cobigen/javaplugin/unittest/merger/JavaMergerTest.java
+++ b/cobigen/cobigen-javaplugin-parent/cobigen-javaplugin/src/test/java/com/devonfw/cobigen/javaplugin/unittest/merger/JavaMergerTest.java
@@ -366,7 +366,8 @@ public class JavaMergerTest {
 
         for (JavaMethod method : resultClazz.getMethods()) {
             JavaMethod origMethod = origClazz.getMethodBySignature(method.getName(), method.getParameterTypes());
-            assertThat(method.getCodeBlock()).isEqualTo(origMethod.getCodeBlock());
+      assertThat(method.getCodeBlock().replaceAll("\r\n|\r|\n", ""))
+          .isEqualTo(origMethod.getCodeBlock().replaceAll("\r\n|\r|\n", ""));
         }
     }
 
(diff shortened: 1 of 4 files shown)


sujith-mn:
> There are some issues with the PR
> 
> 1. as it has multiple changes and it's pushed against master.
> 2. I don't see these issues fixing anything as on master all tests are running fine. Why did you change it? Could you give a pointer on the issue you think, which is solved by this?

JUNIT tests cases were failing, may be the issues are happening only while running locally.    Identified it as due to the newline character from the resource text files.

maybeec:
Could you please give an example which test is failing on your local and which windows you are using? Just want to understand the behavior.

sujith-mn:
Windows 10 is used for the testing.

PropertyMergerTest.testPropertyMergeOverride
The below assertion is failing 

 Assert.assertFalse("nachNameOriginal", mergedPropFile.contains("nachNameOriginal"));
 
Error message:

java.lang.AssertionError: nachNameOriginal
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at com.devonfw.cobigen.propertyplugin.PropertyMergerTest.testPropertyMergeOverride(PropertyMergerTest.java:36)

Issue at code:

  for (String patchLine : patch.split(LINE_SEPARATOR)) {  

Here splitting the line using the current OS specific line separator, but the file to be splitted might have created on a different OS.

maybeec:
That's interesting. Is there somebody, who can also please check that these tests are failing on Win10? I cannot see them failing at least on Win7. @jdiazgon could you please check on Win10?
On travis it does not seem to fail even. So I doubt, this is a fix of anything generally failing. In this case, we have to check, what's different on @sujithmn PC.

jdiazgon:
Hi @maybeec and @sujith-mn,

Interestingly, it seems that for me it is also passing the tests. When I created issue #1083 the test was failing, maybe something in my local was wrong related to line endings. But now it seems to work fine.

Therefore I would recommend to leave it as it is and not merge these changes.

maybeec:
Ok, even the travis jobs are running fine now, then I will close it.
Even, I did not see any of such issue at the users so far.

PR #1133, opened 2020-04-08
Bugfix/1125 generation of templates from external increments does not work

### LINKED ISSUE #1125: Generation of templates from external increments does not work
### Expected behavior

Templates from external increments should be found and generated

### Actual behavior

No templates from external increments can be found if the trigger which defines the increments is not matched

Same as in #747 without variableAssignment 

### Related/Dependent Issues
#747 

### Comments/Hints:
Do not try to define variables in the trigger. The problem with variableAssignment will be fixed later on in #747

lilliCao:
Fixes #1125 .

Implements

* Fix the test for generating templates from external increments
* Adding templates from DependentIncrement to the list of template in TemplatesConfiguration

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/java/com/devonfw/cobigen/systemtest/GenerationTest.java
+++ b/cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/java/com/devonfw/cobigen/systemtest/GenerationTest.java
@@ -2,6 +2,7 @@ package com.devonfw.cobigen.systemtest;
 
 import static com.devonfw.cobigen.test.assertj.CobiGenAsserts.assertThat;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.sameInstance;
 import static org.mockito.Matchers.any;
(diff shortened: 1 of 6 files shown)


cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/java/com/devonfw/cobigen/systemtest/GenerationTest.java
```diff
@@ -90,21 +90,25 @@ public void testGenerationWithExternalIncrements() throws Exception {
...
 
-        // assert
-        IncrementTo externalIncrement = null;
         // we try to get an increment containing external increments
         for (IncrementTo inc : increments) {
             if (inc.getId().equals("3")) {
-                externalIncrement = inc;
+                // We expect increment 3 to have an external increment 0 containing one template
+                assertThat(inc).isNotNull();
+                assertThat(inc.getDependentIncrements().get(0).getTemplates().size()).isEqualTo(1);
```
maybeec, line 98: Please use assertJ assertions to provide better error messages on failing tests.
Here you could use assertThat(Inc.getDependentIncrements().get(0).getTemplates()).hasSize(1)

cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/java/com/devonfw/cobigen/systemtest/GenerationTest.java
```diff
@@ -90,21 +90,25 @@ public void testGenerationWithExternalIncrements() throws Exception {
...
+            }
+            if (inc.getId().equals("4") || inc.getId().equals("5")) {
+                // We expect increment 4 or 5 to have an external increment 0 containing 4 templates
+                assertThat(inc).isNotNull();
+                assertThat(inc.getDependentIncrements().get(0).getTemplates().size()).isEqualTo(4);
             }
         }
```
maybeec, line 105: Be careful, you never assert that templates 3,4,5 are even existing in the list

cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/resources/testdata/systemtest/GenerationTest/externalIncrementsGeneration/context.xml
```diff
@@ -5,7 +5,7 @@
         </matcher>
     </trigger>
     <trigger id="valid_increment_composition" type="mockplugin" templateFolder="valid_increment_composition">
-        <matcher type="fqn" value=".+">
+        <matcher type="unvalid" value="What ever is still unmatched">
```
maybeec, line 8: typo: "invalid"

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/TemplatesConfiguration.java
```diff
@@ -70,6 +70,14 @@ public TemplatesConfiguration(Path configRoot, Trigger trigger, ConfigurationHol
...
         templates = reader.loadTemplates(trigger);
         increments = reader.loadIncrements(templates, trigger);
+        for (Increment inc : increments.values()) {
+            for (Increment incDep : inc.getDependentIncrements()) {
+                for (Template template : incDep.getTemplates()) {
+                    templates.put(template.getName(), template);
+                    inc.getTemplates().add(template);
+                }
+            }
+        }
```
maybeec, line 80: the architecture is the following: The templates configration just holds the configuration and the reader reads information and constructs and collects all information. So it would be better, if you can move this code into the reader rather than here to not mix things up.

cobigen/cobigen-core-parent/cobigen-core-systemtest/src/test/java/com/devonfw/cobigen/systemtest/GenerationTest.java
maybeec: even better, in such a case you could use nice assertJ syntax for that like:

```java
assertThat(inc)
  .extracting(e -> e.getId(), e -> e.getDependentIncrements().get(0).getTemplates().size())
  .containsExactlyInAnyOrder(tuple("3", 1), tuple("4", 4), tuple("5", 4));
```

This will automatically check the existence of exactly three elements having exactly the properties you mentioned here.
There are alos other functions like containsExactly or contains depending on what you want to test.

PR #1127, opened 2020-04-06
#1119 added new TemplatesUtilsClassesUtil class to core (needs to be …

jan-vcapgemini:
…used by eclipse-plugin, maven-plugin and cobigen-cli)

Adresses/Fixes #1119 #1110 .

Implements

* new centralized ResolveUtilClasses handling
* 
* 

@devonfw/cobigen

PR #1119, opened 2020-03-18
(Maven-Plugin) Resolve Template utility classes from core

jan-vcapgemini:
Adresses #1110 #1119 #1127 .

Depends on: #1127

added more debug log messages
added new getContextConfiguration method (returns URL of context.xml)
replaced fixed Artifact file detection with getContextConfiguration method
replaced contextConfiguration and templateRoot logic with new getContextConfiguration method

@devonfw/cobigen

PR #1136, opened 2020-04-14
(Cobigen-CLI) Resolve Template utility classes from core

jan-vcapgemini:
cleaned up CLI CobiGenUtils class (removed getTemplates and resolveUtilClassesFromJar methods)
adjusted generate method to new core (replaced utilClasses with classLoader)
adjusted tsplugin dependency to 2.4.2

Adresses/Fixes #1127 #1119 .

Depends on: #1127

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
+++ b/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
@@ -124,9 +124,12 @@ public class GenerateCommand implements Callable<Integer> {
                 if (inputFiles.size() > 1) {
                     finalTemplates = toTemplateTo(preprocess(cg, TemplateTo.class));
                 }
+
+                ClassLoader inputClassLoader = getClass().getClassLoader();
+
                 for (File inputFile : inputFiles) {
-                    generate(inputFile, ParsingUtils.getProjectRoot(inputFile), finalTemplates, cg,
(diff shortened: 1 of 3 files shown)


cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -124,19 +124,23 @@ public Integer call() throws Exception {
                 if (inputFiles.size() > 1) {
                     finalTemplates = toTemplateTo(preprocess(cg, TemplateTo.class));
                 }
+
+                ClassLoader inputClassLoader = getClass().getClassLoader();
```
jdiazgon, line 128: Have you tested this with the CobiGen CLI fat jar? What I mean is:

1. Run `mvn clean install` on the CLI. That will create a CobiGen.jar inside target folder (our users will install that).
2. Execute `cobigen.bat` or either open a console and execute `java -javaagent:"%~dp0\class-loader-agent.jar" -jar "%~dp0\cobigen.jar" %*` where in `%*` you need to put the correct arguments like `generate EmployeeEntity.java`

jan-vcapgemini: Tested and seems to work.

jdiazgon: Great, thanks!

PR #1163, opened 2020-05-07
Fix for template generation issue 1161

### LINKED ISSUE #1161: Cobigen failed to generate code after doing adapt templates and updating templates
### Expected behavior

Expected to generate code for the selected resources after adapting and updating templates. 

### Actual behavior

Fails with the below error mentioned in the comment

1.  create an entity class as mentioned below
2.  right click -> cobigen -> adapt templates -> generate .

### Comments/Hints:
##Input:
public class EmployeeEntity {
  private String name;
  private Integer age;
  private String location;
}

##Selected Resources:
1. CRUD Rest service
2. CRUD SpringData Repository
3. CRUD UC logic
4. Entity Infrastructure
5. TOs

### Affected version:
* devon-ide version: 2020.04.001
* OS: Windows
* Browser: Chrome

isandesh1986:
Adresses/Fixes #1161 .

Implements

Fix for template generation issue 1161
* 
* 

@devonfw/cobigen

### DIFF
--- a/cobigen-templates/templates-devon4j/src/main/java/com/devonfw/cobigen/templates/devon4j/utils/JavaUtil.java
+++ b/cobigen-templates/templates-devon4j/src/main/java/com/devonfw/cobigen/templates/devon4j/utils/JavaUtil.java
@@ -389,9 +389,7 @@ public class JavaUtil {
      * @return Return type of the method annotated with the given annotation, else "null"
      * @throws ClassNotFoundException 
      */
-    @SuppressWarnings("unchecked")
-   	public String getReturnTypeOfMethodAnnotatedWith(Class<?> pojoClass, String annotatedClassName) throws ClassNotFoundException {
-       	Class<? extends Annotation> classObj = (Class<? extends Annotation>)Class.forName(annotatedClassName);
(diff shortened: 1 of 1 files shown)


maybeec:
Hi Sandesh, be careful as you are changing the logic of the method from testing for full qualified type names to simple type names. Other templates will call this method with the full qualified name and they will fail afterwards.
Couldn't you simply call the method with the full qualified name from the template?

isandesh1986:
> Hi @isandesh1986 , be careful as you are changing the logic of the method from testing for full qualified type names to simple type names. Other templates will call this method with the full qualified name and they will fail afterwards.
> Couldn't you simply call the method with the full qualified name from the template?

Hi @maybeec , All the templates are passing fully qualified type name as string "javax.persistence.EmbeddedId". So we should be good. Let me know.

maybeec:
There seem to be 6 template tests failing. Could you check for the reason?

maybeec:
closing in favor of fix done in d24d6aa00a7f4a9ea90303a1d35086ba2991ca0f
The fix in the PR circumvents class loading, but actually the issue was, that the class was not in classpath

Issue #1170, opened 2020-05-12
CLI option for increment not working


hohwille:
I tried cobigen-cli option `-i` to specify an increment:
```
  -i, --increments=<increments>[,<increments>...]
                       List of increments that will be generated. They need to be
                         specified with numbers separated by comma.
```
As nothing is further specified, I assume that the increment has to be specified by its `name` attribute from `templates.xml`.
I therefore supplied that value via Copy&Paste (to ensure no typo happens):
```
    <increment name="tos" description="Common (ETO)">
```
So I called:
```
cg generate -i tos -v app/src/main/java/com/customer/project/app/component/dataaccess/api/FooEntity.java
[DEBUG] Input files and output root path confirmed to be valid.
[INFO ] As you did not specify where the code will be generated, we will use the project of your current Input file.
[DEBUG] Generating to: /Users/hohwille/projects/myproject/workspaces/main/app/core
[INFO ] No increment with that name has been found, Please provide correct increment name and try again ! Thank you
```
If the last line means that generation failed because the specified increment `tos` could not be found, I would expect this to be an ERROR and not an INFO.

hohwille:
BTW: If there is only one single increment available it is rather rediculous that CobiGen CLI is asking wheter to generate 
* all
* the only available increment.

Further, when CobiGen CLI is printing the available increments, it should also print the technical identifiers (`name` attribute) in brackets so it is easier to figure the option parameter for `-i` (if the feature is working after all).

jdiazgon:
Hi @hohwille,

CobiGen uses the "Description" of the increments in order to search for them. Descriptions is what our users will see on the Eclipse UI. Name is just used internally for configuration of the increments, so it does not make much sense to be used.

![image](https://user-images.githubusercontent.com/31688036/81780339-aa678c80-94f6-11ea-8c46-b7c43870e42e.png)

isandesh1986:
Yes, just verified. cli is searching based on the description.

<img width="947" alt="tos" src="https://user-images.githubusercontent.com/63033584/81938354-6a95c780-9612-11ea-9a52-be61f1ee68e2.png">

<img width="949" alt="rest" src="https://user-images.githubusercontent.com/63033584/81938375-71243f00-9612-11ea-8483-6df6706c533b.png">

hohwille:
[...]

PR #1184, opened 2020-05-21
CLI increments to search based on both name and description of the

isandesh1986:
templates #1170

Adresses/Fixes #<ISSUENO>.

Implements

* 
* 
* 

@devonfw/cobigen


maybeec:
Thanks for the new pr, looks fine. Could you please adapt the documentation accordingly?

maybeec:
I think you have to remove the threshold test on equals comparison. Otherwise, generation will even start for similar but not equal strings. Am I right?

isandesh1986:
> I think you have to remove the threshold test on equals comparison. Otherwise, generation will even start for similar but not equal strings. Am I right?

Right. My bad! I missed to remove it. Have taken care now. Kindly check.

isandesh1986:
> Thanks for the new pr, looks fine. Could you please adapt the documentation accordingly?

Documentation also has been updated. Thanks

PR #1188, opened 2020-06-08
Bugfix/1158 external servers does not update properly

### LINKED ISSUE #1158: tsplugin nestserver update not working
### Expected behavior

As a user, I want to update tsplugin so that I can use the new features, bugfixes.

### Actual behavior

Updating the tsplugin seems to work, but my bugs of the ts-merger are not resolved.

1. install version x of the ts-plugin in your eclipse
2. run once a generation which will trigger the ts-merger (the nestserver will be downloaded)
3. install a new version x+1 of the ts-plugin, which includes changes in the ts-merger
4. re-run the generation and check whether the fixed behavior of the ts-plugin is resolved

### Comments/Hints:

The problem occurs as the nestserver is stored without any information of the version into the home folder. The logic of the ts-plugin simply checks the nestserver to exist and if so, will not do anything, but simply run it. In case a new ts-plugin version will declare a new version of the nestserver, this will never be taken up as the binary is never updated if existing.

In addition, please respect the requirement, that there should be the ability to maintain multiple versions of the nestserver as you might have multiple installations of the ts-plugin on your machine in different versions. Each of the plugin should get its corresponding nestserver version on start.

### Affected version:

* OS: Windows/Linux/Mac?
* Browser: Chrome/Firefox/Safari?

lilliCao:
Fixes #1158 

Implements

* Do not remove old external servers
* fix acquirePort in initializeConnection() by calling startConnection() again to have new port in url 
* Add check if the correct version is currently deployed

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
+++ b/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
@@ -493,17 +493,21 @@ public class ExternalProcessHandler {
             if (conn.getResponseCode() < 300) {
 
                 // Check if it is the correct server version
-                BufferedReader br = new BufferedReader(new InputStreamReader(conn.getInputStream()));
-                String response = br.readLine();
+                try (InputStreamReader isr = new InputStreamReader(conn.getInputStream());
+                    BufferedReader br = new BufferedReader(isr)) {
+                    String response = br.readLine();
 
-                if (response.equals(processProperties.getServerVersion())) {
-                    return false;
-                }
+                    if (response.equals(processProperties.getServerVersion())) {
+                        return false;
+                    }
 
-                if (response.equals("true")) {
-                    LOG.warn(
-                        "An old version is currently deployed. Please consider deploying the newest version to get the current bug fixes/features");
-                    return false;
+                    if (response.equals("true")) {
+                        LOG.warn(
+                            "An old version is currently deployed. Please consider deploying the newest version to get the current bug fixes/features");
+                        return false;
+                    }
+                } catch (Exception e) {
+                    LOG.info("Reading server version failed");
                 }
 
                 return true;
--- a/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
+++ b/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
@@ -1,10 +1,12 @@
 package com.devonfw.cobigen.impl.externalprocess;
 
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.net.HttpURLConnection;
@@ -407,6 +409,8 @@ public class ExternalProcessHandler {
                 // Just check correct port acquisition
                 if (acquirePort()) {
                     return true;
+                } else {
+                    continue;
                 }
             } catch (Exception e) {
                 LOG.error("Connection to server failed, attempt number " + retry + ".");
@@ -487,7 +491,22 @@ public class ExternalProcessHandler {
         try {
             getConnection("GET", "Content-Type", "text/plain", ExternalProcessConstants.IS_CONNECTION_READY);
             if (conn.getResponseCode() < 300) {
-                return false;
+
+                // Check if it is the correct server version
+                BufferedReader br = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+                String response = br.readLine();
+
+                if (response.equals(processProperties.getServerVersion())) {
+                    return false;
+                }
+
+                if (response.equals("true")) {
+                    LOG.warn(
+                        "An old version is currently deployed. Please consider deploying the newest version to get the current bug fixes/features");
+                    return false;
+                }
+
+                return true;
             }
         } catch (IOException e) {
             LOG.error("Connection to server failed, maybe the server is not yet deployed...");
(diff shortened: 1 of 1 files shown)


cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
```diff
@@ -445,6 +409,8 @@ public boolean initializeConnection() {
                 // Just check correct port acquisition
                 if (acquirePort()) {
                     return true;
+                } else {
+                    continue;
```
maybeec, line 413: it's ok, but actually not needed as the look will anyhow continue if you don't return from the function.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
```diff
@@ -525,7 +491,22 @@ public boolean isNotConnected() {
...
+                BufferedReader br = new BufferedReader(new InputStreamReader(conn.getInputStream()));
+                String response = br.readLine();
+
+                if (response.equals(processProperties.getServerVersion())) {
+                    return false;
+                }
+
+                if (response.equals("true")) {
+                    LOG.warn(
+                        "An old version is currently deployed. Please consider deploying the newest version to get the current bug fixes/features");
```
maybeec, line 505: Exactly this should be done automatically.
There is a download function at the moment, which automatically download the nestserver, but does not store it as far as I know including the version number.
This should be kept. So in case a previous version of the tsplugin for example downloads nestserver-1.0.5.exe and then you run a new instance of the tsplugin, which depends on nestserver-1.0.7.exe. In this case automatically the new version should be downloaded as well and should automatically be run.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
lilliCao: oh yeah. I muss oversaw it

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
lilliCao: Let me explain the logic here briefly. E.g the tsplugin wants to use nestserver. The first thing it does is check if it can connect to a particular port where the server should be running (in this case 5000). If there is an old version of nestserver running on that port, it will use that version and does not trigger any download. As that the new version will not get downloaded at all. Also even if it is somehow got downloaded to the local folder, it will never be started.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
```diff
@@ -525,7 +491,22 @@ public boolean isNotConnected() {
         try {
             getConnection("GET", "Content-Type", "text/plain", ExternalProcessConstants.IS_CONNECTION_READY);
             if (conn.getResponseCode() < 300) {
-                return false;
+
+                // Check if it is the correct server version
+                BufferedReader br = new BufferedReader(new InputStreamReader(conn.getInputStream()));
```
maybeec, line 496: This is a memory leak. you should initialize these readers in a try-with-resource construct to make them close afterwards.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
maybeec: ok, smart. And in case it's not the correct version, would cobigen try to download automatically the new version?
This would be necessary for the fix for me.

lilliCao: Exactly. In case cobigen finds out that it is connecting to the wrong server version, it will download the version defined in pom and start it

maybeec: great. One last question: there was a time, when the nestserver exe was not downloaded and stored including a version suffix. Meaning it was downloading for example nestserver-1.0.0.exe from the web, but simply stored nestserver.exe on the filesystem, which would break the ability to have multiple versions of the nestserver next to each other.
Could you once please check that this is not the case?

maybeec:
Could you please try to test this bugfix? Would be nice to be sure in future, that it will not come again.
In general we should anyhow do it with all bugfixes to have a test reproducing it.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/externalprocess/ExternalProcessHandler.java
lilliCao: The nestserver exe should always contains the version. If it is not, it's properly a broken release of nestserver itself. Cobigen does not rename the downloaded file. The reason why there are no multiple versions of nestserver in the past is because in the old logic, cobigen cleans up all the old versions of nestserver if it downloads a new one.

lilliCao:
I did test the combination of these cases manually: (no wanted-version at local, wanted-version at local but not running, wanted-version at local and running) with ( no not-wanted-version at local, not-wanted-version at local but not running, not-wanted-version at local and running). It works with all the version of nestserver which modifies the /isConnectionReady endpoint to return server version (>=1.1.7). Nestserver v<=1.1.5 does not return the server version but return "true". And there is no really good way to distinguish among them, cobigen can work with those versions too and will leave a warning to let the user know and update the nestserver version to the newest one.

maybeec:
is there the ability to test it also automatically?

lilliCao:
I am not sure. I find it difficult because it requires many test cases with different version of nestservers and different state of their existence/running mode at local machine

Issue #1169, opened 2020-05-12
CobiGen has no consistent template conventions, custom templates not working, NPEs


hohwille:
### Expected behavior

As a CobiGen user, I want to use my custom templates so that I can generate code specific for the needs of my projects.

### Actual behavior

CobiGen has no clearly designed and documented concept for templates and the template detection is full of hacks, inconsistencies and bugs:
* https://github.com/devonfw/cobigen/blob/master/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/utils/CobiGenUtils.java#L381
* https://github.com/devonfw/cobigen/blob/e296465791895f63a16618612127a25511849289/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/TemplatesJarUtil.java#L29
* https://github.com/devonfw/cobigen/blob/4942eede166552fab4479faa617a0d1bf1b5cb0c/cobigen/cobigen-core-parent/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/constants/TemplatesJarConstants.java#L16

I get NPEs like:
```
java.lang.NullPointerException
at com.devonfw.cobigen.cli.utils.CobiGenUtils.initializeCobiGen(CobiGenUtils.java:89)
at com.devonfw.cobigen.cli.commands.GenerateCommand.call(GenerateCommand.java:123)
at com.devonfw.cobigen.cli.commands.GenerateCommand.call(GenerateCommand.java:1)
at picocli.CommandLine.executeUserObject(CommandLine.java:1680)
at picocli.CommandLine.access$900(CommandLine.java:144)
at picocli.CommandLine$RunLast.handle(CommandLine.java:2038)
at picocli.CommandLine$RunLast.handle(CommandLine.java:2005)
at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:1872)
at picocli.CommandLine.execute(CommandLine.java:1801)
at com.devonfw.cobigen.cli.CobiGenCLI.main(CobiGenCLI.java:43)
```

So to summarize what CobiGen currently demands:
* templates artifact name has to contain `devon4j` for CobiGen-CLI
* templates artifact name has to match `templates-([^-]+)-(\d+\.?)+.jar$`. Side note: I renamed my templates artifactId to `templates-devon4j` but then noticed that `templates-devon4j-2.5.0-snapshot.jar` does not match this regex.
[...]

PR #1222, opened 2020-08-21
Enable cobigen to be created without configFileorFolder passed

lilliCao:
Adresses #1169 .

Implements

* ConfigurationUtils class: read and process configuration file. 
* Implementing the logics for searching for the templates location
   * Search for the configuration file .cobigen which can be located in either 
      *  a directory where the environment variable COBIGEN_CONFIG_DIR is pointed to
      *  cobigen home
   * If no configuration file is found, or no templates location is found in the configuraton file, it then searches in the cobigen home (.cobigen/templates/)
                        
* Add create() method in CobigenFactory to enable create cobigen instance without any given configFileorFolder

@devonfw/cobigen


cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param cobigenConfigFile:
+     *            cobigen configuration file
+     * @return Properties containing configuration
+     */
+    private static Properties readConfigrationFile(Path cobigenConfigFile) {
+        Properties props = new Properties();
+        try (BufferedReader reader = Files.newBufferedReader(cobigenConfigFile, Charset.forName("UTF-8"))) {
+            props.load(reader);
+        } catch (IOException e) {
+            LOG.error("An error occured while reading the config file", e);
```
maybeec, line 84: please print out the path which could be read here as part of the log.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Find the cobigen configuration file
+     * @return a path of cobigen configuration file if exist, otherwise null
+     */
+    private static Path getConfigurationFile() {
+        String envValue = System.getenv(ConfigurationConstants.COBIGEN_CONFIG_DIR);
+        if (StringUtils.isNotEmpty(envValue)) {
+            return Paths.get(envValue).resolve(ConfigurationConstants.COBIGEN_CONFIG_FILE);
```
maybeec, line 63: Implement a fallback in case the .cobigen file does not exist in the cobigen config dir and continue trying the cobigenHome. Currently you don't check for the file to exist and will never check for a configuration file in home.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private static final Logger LOG = Logger.getLogger(ConfigurationUtil.class);
+
+    /**
+     * The method finds location of templates. It could be CobiGen_Templates folder or a template artifact
+     * @return template location uri if exist, otherwise null
+     */
+    public static URI findTemplatesLocation() {
+        Path configFile = getConfigurationFile();
+        if (configFile != null && Files.exists(configFile)) {
```
maybeec, line 34: you should better simply return null by #getConfigurationFile() in case no file exists, then you can implement the fallback within the method and don't need to check for existence again here.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+            LOG.info("Configuration file is located at " + configFile);
```
maybeec, line 35: if you write log statements with additional data to be printed, please use log4j stringbuilder notation. In this case: `             LOG.info("Configuration file is located at {}", configFile);`
The difference here is, that string concatenation is already done before entering the info() method and thus this is lost computation power in case no INFO log should be written. By slf4j string builder notation, no computation is done at all before the decision is made by slf4j to print the log statement.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+                Path templatesPath = Paths.get(templatesLocation);
+                if (Files.exists(templatesPath)) {
+                    return Paths.get(templatesLocation).toUri();
+                } else {
+                    LOG.info(
+                        "Could not find the given templates location in configuration file. Try to look for templates in cobigen home");
+                }
+            } else {
+                LOG.info(
+                    "No templates location is found in configuration file. Try to look for templates in cobigen home");
```
maybeec, line 48: Better indicate the property which is not set to teach the user on how to change the behavior.
For example `LOG.info("Property {} not set in {}. Trying to look for templates in cobigen home directory {}", ConfigurationConstants.COBIGEN_CONFIG_TEMPLATES_LOCATION_KEY, configFile, <variable indicating cobigen home dir>);`

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+            LOG.info("Configuration file is located at " + configFile);
+            Properties props = readConfigrationFile(configFile);
+            String templatesLocation = props.getProperty(ConfigurationConstants.COBIGEN_CONFIG_TEMPLATES_LOCATION_KEY);
+            if (StringUtils.isNotEmpty(templatesLocation)) {
+                Path templatesPath = Paths.get(templatesLocation);
+                if (Files.exists(templatesPath)) {
+                    return Paths.get(templatesLocation).toUri();
+                } else {
+                    LOG.info(
+                        "Could not find the given templates location in configuration file. Try to look for templates in cobigen home");
```
maybeec, line 44: same here as below. Try to give more information to the user.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+                }
+            } else {
+                LOG.info(
+                    "No templates location is found in configuration file. Try to look for templates in cobigen home");
+            }
+        } else {
+            LOG.info("No configuration file is found. Try to look for templates in cobigen home");
```
maybeec, line 51: same here.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
```diff
@@ -0,0 +1,112 @@
...
+            // use Cobigen_Templates folder
+            return templatesFolderPath.toUri();
+        } else {
+            // use template jar
+            File templateJar = TemplatesJarUtil.getJarFile(false, templatesPath.toFile());
+            if (templateJar != null && Files.exists(templatesPath)) {
+                return templateJar.toURI();
+            }
+        }
+        LOG.info("Could not find any templates in cobigen home");
```
maybeec, line 109: Please provide some more information to the user.

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/CobiGenFactory.java
```diff
@@ -51,6 +52,20 @@ public static CobiGen create(URI configFileOrFolder) throws InvalidConfiguration
...
+     * Creates a new {@link CobiGen}
+     *
+     * @return a new instance of {@link CobiGen}
+     * @throws IOException
+     *             if the {@link URI} points to a file or folder, which could not be read.
+     * @throws InvalidConfigurationException
+     *             if the context configuration could not be read properly.
+     */
+    public static CobiGen create() throws InvalidConfigurationException, IOException {
+        URI configFileOrFolder = ConfigurationUtil.findTemplatesLocation();
```
maybeec, line 65: I think we should provide a proper error message in case no templates / configuration folder could be found as the error message in `public static CobiGen create(URI configFileOrFolder)` is not providing a proper message in case of null parameter. Maybe we can change the error message to some more meaningful which would also give a CLI user proper knowledge of what has happened.

maybeec (changes requested):

cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationUtil.java
lilliCao: mmm I do try to search for it in the cobigenHome as well. Only in case there is no config file found (in the place pointed by the env variable and in cobigen home), null is returned

MikeSchumacherCapgemini:
Have you checked that there are tests that would fail if this feature does not work correctly? Otherwise you should add a test/tests that check if the implemented functionality works when it is supposed to and doesn't work when it's not supposed to.

lilliCao:
There should be no tests which will be effected by this feature. Yeah I can add some to test the functionalities.

maybeec (approved):

PR #1295, opened 2020-12-11
Update core dependencies in eclipse plugin

MikeSchumacherCapgemini:
Updated core dependency to fit the latest release


maybeec:
missing the update of the plugin xml and build properties we have done in the screenshare

PR #1300, opened 2020-12-17
Include accumulationType to hashCode

### LINKED ISSUE #1299: Matcher with identical type and value is evaluated wrong
### Expected behavior

I have a package with two types of classes: The first type extends a base class and the other don't extend. For those two classes I would like to write two different triggers:
```xml
 <trigger id="ts_enums" type="java" templateFolder="ts_enums">
     <matcher type="expression" value="instanceof com.test.BaseEnum" accumulationType="AND" />
  </trigger>
   
  <trigger id="ts_datatypes" type="java" templateFolder="ts_datatypes">
     <matcher type="expression" value="instanceof com.test.BaseEnum" accumulationType="NOT" />
  </trigger>
```
The only difference in those matchers is the accumulation type.

### Actual behavior
I executed the cobigen maven plugin for a single file which matches the second trigger but the generation is also executed for the first trigger. 

1. Write the triggers from above.
2. Generate for a classes which matches only to one matcher.

### Comments/Hints:
The problems seems to be a missing implementation of hash code in com.devonfw.cobigen.impl.config.entity.Matcher. The AbstractMatcher implements hash code for type and value but the Matcher subclass doesn't. In the generation the com.devonfw.cobigen.impl.aop.CachedInterceptor is used for caching and this caching depends on proper hash code implementation of the method parameters. As the accumulation type is not used in the hash code implementation the hash code for the above two matchers returns the same value even though the accumulation type is different.

### Affected version:
cobigen-maven-plugin 7.0.0

dachiaradia:
Adresses/Fixes #1299 .

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/entity/Matcher.java
+++ b/cobigen/cobigen-core-parent/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/entity/Matcher.java
@@ -2,6 +2,7 @@ package com.devonfw.cobigen.impl.config.entity;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 
 import com.devonfw.cobigen.impl.config.entity.io.AccumulationType;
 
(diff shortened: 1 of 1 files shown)


maybeec (approved):

maybeec (dismissed):

maybeec:
@danielchiaradia, we need your sign of the CLA to be able to merge this PR

danielchiaradia:
@maybeec Done

maybeec:
@ghost finally done by #1485 
There was some disruption with refactorings but finally got it into the next release. Sorry for the tremendous delay

PR #1363, opened 2021-07-01
check for name instead of Class

LarsReinken:
Adresses/Fixes 
Fixes an issue in which an openAPI increment wasn't shown, when using the containerMatcher in combination with the EntityNameRegexMatcher


maybeec (approved):
Temporary fix as of urgency, @LarsReinken, need to have a test for that and please re-check whether the old solution can be restored and fixed by proper class loading.
Please create a new issue out of that to fix it properly.

PR #1366, opened 2021-07-09
Open api regex trigger fix

LarsReinken:
added a test for the regexTriggerMatcher


MikeSchumacherCapgemini (dismissed):

maybeec (changes requested):
Please make sure, that this test is properly testing the fix or whether the fix is not needed anymore. I tried, but could not see red-green effect of the new test when reverting the "fix" and reapplying it.

LarsReinken:
> Please make sure, that this test is properly testing the fix or whether the fix is not needed anymore. I tried, but could not see red-green effect of the new test when reverting the "fix" and reapplying it.

I came to the same result, the matcher seems to work without any change.

PR #1390, opened 2021-08-23
#1372 fixed META-INF output directory

### LINKED ISSUE #1372: OutOfMemoryError while adapting the template in Eclipse
Getting the below error while adapting the template.

image

Heap size used:  -Xms256m-Xmx2048m

jan-vcapgemini:
Adresses/Fixes #1372 .

@devonfw/cobigen

### DIFF
--- a/cobigen-templates/pom.xml
+++ b/cobigen-templates/pom.xml
@@ -32,8 +32,7 @@
         <targetPath>src/main/templates</targetPath>
       </resource>
       <resource>
-        <directory>src/main/resources/META-INF</directory>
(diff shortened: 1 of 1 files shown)


MikeSchumacherCapgemini (approved):

MikeSchumacherCapgemini:
Using ./ as the target path led to a folder with the name "." being created.
This is a workaround since we could not find a way to properly target the jars root directory (since ./ or similar did not work)

maybeec (approved):

Issue #1389, opened 2021-08-19
Generate from devonfw example project with CLI is not working


jan-vcapgemini:
### Expected behavior

As a CobiGen CLI user, I want to generate artifacts from the devonfw example project using a simple CustomerEntity.

### Actual behavior
When running generate on the example project, following error occurs:

`Running: cobigen generate cobigenexample/core/src/main/java/com/example/application/cobigenexample/customermanagement/dataaccess/api/CustomerEntity.java
[DEBUG] Input files and output root path confirmed to be valid.
[ERROR] Invalid input for CobiGen, please check your input file 'cobigenexample/core/src/main/java/com/example/application/cobigenexample/customermanagement/dataaccess/api/CustomerEntity.java'
[ERROR] There are no common Templates/Increments which could be generated from every of your inputs. Please think about executing generation one by one input file.
java.util.InputMismatchException: No compatible input files.
        at com.devonfw.cobigen.cli.commands.GenerateCommand.preprocess(GenerateCommand.java:177)
        at com.devonfw.cobigen.cli.commands.GenerateCommand.doAction(GenerateCommand.java:115)
        at com.devonfw.cobigen.cli.commands.CommandCommons.call(CommandCommons.java:36)
        at com.devonfw.cobigen.cli.commands.CommandCommons.call(CommandCommons.java:15)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
        at picocli.CommandLine.access$1100(CommandLine.java:145)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
        at picocli.CommandLine.execute(CommandLine.java:2058)
        at com.devonfw.cobigen.cli.CobiGenCLI.main(CobiGenCLI.java:37)
[INFO ] Failed`

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

Same steps can be easily followed here: https://www.katacoda.com/devonfw-dev/scenarios/170_remove-adapttemplatescobigen_cobigen

1. Create a fresh devonfw ide project
2. Run: devon cobigen
3. Switch to your workspaces/main directory
4. Run: devon java create com.example.application.cobigenexample
[...]

PR #1425, opened 2021-11-16
#1389 added test to check if a version using dependencyManagement can…

jan-vcapgemini:
Added test to check if a version using dependencyManagement can be resolved from a parent project (especially if a maven.config is present)

Adresses #1389.

This PR is waiting for https://github.com/m-m-m/code/issues/35 to be resolved first.
The m-m-m.code version needs to be adjusted for this test to succeed.

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/cli/src/main/resources/pom.xml
+++ b/cobigen-cli/cli/src/main/resources/pom.xml
@@ -64,17 +64,17 @@
     <dependency>
       <groupId>net.sf.m-m-m</groupId>
       <artifactId>mmm-code-java-parser</artifactId>
-      <version>1.0.0-beta6</version>
+      <version>1.0.0-beta7</version>
     </dependency>
     <dependency>
       <groupId>net.sf.m-m-m</groupId>
       <artifactId>mmm-code-base</artifactId>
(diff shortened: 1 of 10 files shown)


cobigen-cli/cli/src/main/resources/pom.xml
```diff
@@ -64,17 +64,17 @@
...
+      <version>1.0.0-beta7</version>
     </dependency>
     <dependency>
       <groupId>net.sf.m-m-m</groupId>
       <artifactId>mmm-code-java-maven</artifactId>
-      <version>1.0.0-beta6</version>
+      <version>1.0.0-beta7</version>
     </dependency>
     <dependency>
       <groupId>net.sf.m-m-m</groupId>
```
maybeec, line 80: why do we manage mmm dependencies here as well? sounds really strange

maybeec (changes requested):

cobigen-cli/cli/src/main/resources/pom.xml
jan-vcapgemini: I've removed them now.

maybeec (approved):

PR #1445, opened 2021-12-15
#1434 added repeatable prompt for output path of OpenApi generation

### LINKED ISSUE #1434: CobiGen CLI generate without output path does not work without a pom.xml
### Expected behavior

As a CobiGen CLI user, I want to generate from an OpenAPI file without having to provide a pom in my project.

### Actual behavior
When generating without providing an output path, the generate process fails with a NPE.

```
[INFO ] As you did not specify where the code will be generated, we will use the project of your current Input file.
java.lang.NullPointerException
        at com.devonfw.cobigen.cli.commands.GenerateCommand.setOutputRootPath(GenerateCommand.java:297)
        at com.devonfw.cobigen.cli.commands.GenerateCommand.generate(GenerateCommand.java:264)
        at com.devonfw.cobigen.cli.commands.GenerateCommand.doAction(GenerateCommand.java:117)
        at com.devonfw.cobigen.cli.commands.CommandCommons.call(CommandCommons.java:36)
        at com.devonfw.cobigen.cli.commands.CommandCommons.call(CommandCommons.java:15)
        at picocli.CommandLine.executeUserObject(CommandLine.java:1933)
        at picocli.CommandLine.access$1100(CommandLine.java:145)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2332)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2326)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2291)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2159)
        at picocli.CommandLine.execute(CommandLine.java:2058)
        at com.devonfw.cobigen.cli.CobiGenCLI.main(CobiGenCLI.java:37)

```

1. Create a new empty project folder in your devon-ide workspace
2. Place the example OpenAPI [devonfw.yml](https://github.com/devonfw/cobigen/wiki/files/devonfw.yml) into your project folder
3. Open a CLI window in the current path
4. Run: `cobigen generate devonfw.yml`

### Comments/Hints:
- Providing `-o ./` works fine.
- Could be fixed with another prompt for an output directory

### Affected version:

* CobiGen CLI 2021.08.002

jan-vcapgemini:
Added repeatable prompt for output path of OpenApi generation if automatic detection of output path failed (if user just hits enter, the parent directory of the inputfile gets used as output directory)

set outputRootPath to absolute path in setOutputRootPath method

Adresses/Fixes #1434 .

Implements

* 
* 
* 

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
+++ b/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
@@ -151,8 +151,8 @@ public class GenerateCommand extends CommandCommons {
           && MavenUtil.getProjectRoot(inputFile.toPath(), false) == null) {
 
         LOG.info(
-            "No ouput directory was found. Please specify an ouput path or just press enter to generate your files to this directory: {}",
-            inputFile.getAbsoluteFile().getParent().toString());
+            "No output directory was found. Please specify an output path or just press enter to generate your files to this directory: {}",
+            inputFile.getAbsoluteFile().getParent());
 
         String userInput = getUserInput();
 
--- a/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
+++ b/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
@@ -144,28 +144,23 @@ public class GenerateCommand extends CommandCommons {
       boolean isJavaInput = extension.endsWith(".java");
       boolean isOpenApiInput = extension.endsWith(".yaml") || extension.endsWith(".yml");
 
-      // checks for OpenApi input file, output root path and project root being detectable
-      if (isOpenApiInput && this.outputRootPath == null && MavenUtil.getProjectRoot(inputFile, false) == null) {
+      // checks for output root path and project root being detectable
+      if (this.outputRootPath == null && MavenUtil.getProjectRoot(inputFile, false) == null) {
 
         LOG.info(
-            "No output directory was found. Please specify an output path or just press enter to generate your files to this directory: {}",
-            inputFile.toAbsolutePath().getParent().toString());
+            "Did not detect the input as part of a maven project, the root directory of the maven project was not found.");
 
-        String userInput = getUserInput();
-
-        Path preprocessedFile = Paths.get(userInput);
-
-        if (!userInput.isEmpty()) {
-          while (!ValidationUtils.isOutputRootPathValid(preprocessedFile)) {
-            Path userFolder = Paths.get(getUserInput());
-            preprocessedFile = preprocessInputFile(userFolder);
-          }
-          this.outputRootPath = preprocessedFile;
-          LOG.info("Your output directory was successfully set to: {}",
-              this.outputRootPath.toAbsolutePath().toString());
-        } else {
-          setOutputRootPath(inputFile.toAbsolutePath().getParent());
+        // cancels with an error if more than one input file was provided
+        if (this.inputFiles.size() > 1) {
+          LOG.error(
+              "As multiple input files were provided, setting an automatic output path is not possible. Cancelling...");
+          System.exit(1);
         }
+        LOG.info("Would you like to take '{}' as a root directory for output generation? \n"
+            + "type yes/y to continue or no/n to cancel (or hit return for yes).", System.getProperty("user.dir"));
+
+        setOutputPathWithPrompt();
+
       }
 
       try {
@@ -201,6 +196,37 @@ public class GenerateCommand extends CommandCommons {
         ? generableArtifactSelection(this.increments, toIncrementTo(finalTos), IncrementTo.class)
         : generableArtifactSelection(this.templates, toTemplateTo(finalTos), TemplateTo.class));
     return new Tuple<>(generationInputs, selectedGenerableArtifacts);
+  }
+
+  /**
+   * Opens a repeatable prompt with a yes/no question and sets the output directory to the current user directory.
+   *
+   */
+  private void setOutputPathWithPrompt() {
+
+    Path outputDirectory = Paths.get(System.getProperty("user.dir"));
+    boolean decided = false;
+    while (!decided) {
+      String userInput = getUserInput();
+      switch (userInput) {
+        case "yes":
+        case "y":
+        case "":
+          LOG.info("Set output directory to: {}", outputDirectory.toString());
+          this.outputRootPath = outputDirectory;
+          decided = true;
+          break;
+        case "no":
+        case "n":
+          LOG.info("Cancelling generate process... you can use -o to explicitly set the output directory.");
+          decided = true;
+          System.exit(0);
+          break;
+        default:
+          LOG.info("Invalid input. Please answer yes/n or no/n (or hit return for yes).");
+          break;
+      }
+    }
   }
 
   /**
--- a/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
+++ b/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
@@ -146,6 +146,31 @@ public class GenerateCommand extends CommandCommons {
       boolean isJavaInput = extension.endsWith(".java");
       boolean isOpenApiInput = extension.endsWith(".yaml") || extension.endsWith(".yml");
 
+      // checks for OpenApi input file, output root path and project root being detectable
+      if (isOpenApiInput && this.outputRootPath == null
+          && MavenUtil.getProjectRoot(inputFile.toPath(), false) == null) {
+
+        LOG.info(
+            "No ouput directory was found. Please specify an ouput path or just press enter to generate your files to this directory: {}",
+            inputFile.getAbsoluteFile().getParent().toString());
+
+        String userInput = getUserInput();
+
+        File preprocessedFile = Paths.get(userInput).toFile();
+
+        if (!userInput.isEmpty()) {
+          while (!ValidationUtils.isOutputRootPathValid(preprocessedFile)) {
+            File userFolder = Paths.get(getUserInput()).toFile();
+            preprocessedFile = preprocessInputFile(userFolder);
+          }
+          this.outputRootPath = preprocessedFile;
+          LOG.info("Your output directory was successfully set to: {}",
+              Paths.get(this.outputRootPath.getAbsolutePath()).toString());
+        } else {
+          setOutputRootPath(inputFile.toPath().toAbsolutePath().getParent());
+        }
+      }
+
       try {
         Object input = cg.read(inputFile.toPath(), StandardCharsets.UTF_8);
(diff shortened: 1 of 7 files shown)


maybeec (changes requested):

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -146,6 +146,31 @@ public Integer doAction() throws Exception {
       boolean isJavaInput = extension.endsWith(".java");
       boolean isOpenApiInput = extension.endsWith(".yaml") || extension.endsWith(".yml");
 
+      // checks for OpenApi input file, output root path and project root being detectable
+      if (isOpenApiInput && this.outputRootPath == null
```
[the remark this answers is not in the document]
jan-vcapgemini: Removed the restriction.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -146,6 +146,31 @@ public Integer doAction() throws Exception {
...
+          && MavenUtil.getProjectRoot(inputFile.toPath(), false) == null) {
+
+        LOG.info(
+            "No output directory was found. Please specify an output path or just press enter to generate your files to this directory: {}",
+            inputFile.getAbsoluteFile().getParent());
+
+        String userInput = getUserInput();
```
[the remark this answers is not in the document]
jan-vcapgemini: Could later be improved with #1448

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -294,7 +319,7 @@ private void setOutputRootPath(Path inputProject) {
         "As you did not specify where the code will be generated, we will use the project of your current Input file.");
     LOG.debug("Generating to: {}", inputProject);
 
-    this.outputRootPath = inputProject.toFile();
+    this.outputRootPath = inputProject.toFile().getAbsoluteFile();
```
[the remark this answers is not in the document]
jan-vcapgemini: Adjusted.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -529,15 +550,15 @@ public static String getUserInput() {
...
-    String path = inputFile.getPath();
+    String path = inputFile.toString();
     String pattern = "[\\\"|\\'](.+)[\\\"|\\']";
     boolean matches = path.matches(pattern);
     if (matches) {
       path = path.replace("\"", "");
       path = path.replace("\'", "");
-      return new File(path);
+      return Paths.get(path);
     }
```
maybeec, line 562: I am pretty sure that this code is not needed at all. Could you check removing it? The quoting just comes as some characters need to be escaped which is highly dependent on the actual command line. I would bet that this is handled by picocli as otherwise you would also need to get rid of backslashes here to remove escapes from Unix paths

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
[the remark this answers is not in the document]
maybeec: You are missing the point, we should not try to develop features on top when we could solve it cost efficient and less error prone. I feel putting auto compete for path into scope will raise much complexity as of different OS and so on

maybeec (changes requested):

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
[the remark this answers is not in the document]
jan-vcapgemini: Replaced with yes/no prompt.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: It is still needed, because setTrimQuotes was not able to handle single quotes in CMD.

jan-vcapgemini:
Replaced the prompt for an output directory with a simpler yes/no prompt.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -181,6 +198,37 @@ public Integer doAction() throws Exception {
     return new Tuple<>(generationInputs, selectedGenerableArtifacts);
   }
 
+  /**
+   * Opens a repeatable prompt with a yes/no question and sets the output directory to the current user directory.
+   *
+   */
+  private void setOutputPathWithPrompt() {
+
+    Path outputDirectory = Paths.get(System.getProperty("user.dir"));
```
maybeec, line 207: Why do you want to generate to the user home folder? Sounds crazy. A default should be the path the user is currently navigated while issuing cobigen cli, i.e. pwd

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -181,6 +198,37 @@ public Integer doAction() throws Exception {
...
+        case "":
+          LOG.info("Set output directory to: {}", outputDirectory.toString());
+          this.outputRootPath = outputDirectory;
+          decided = true;
+          break;
+        case "no":
+        case "n":
+          LOG.info("Cancelling generate process... you can use -o to explicitly set the output directory.");
+          decided = true;
+          System.exit(0);
```
maybeec, line 223: Just asking myself whether this makes trouble during testing.
I remembered that I had really issues identifying and maintaining system edits which where coming everywhere in the code. No clear return code facade was done.
If I am not totally wrong, I reimplemented that to have just two exits in the CobiGenCLI top level class or top level command.
Please check. From an architectural point of view, it would be much more convenient to throw a user abort exception here and catch it and edit with a different code to be able to distinguish between an error and user driven abortion in tests.
In devonfw IDE I think we have return code 255 for that. Never googled it but could be a standard and if not let's comply at least with devonfw ide.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -181,6 +198,37 @@ public Integer doAction() throws Exception {
     return new Tuple<>(generationInputs, selectedGenerableArtifacts);
   }
 
+  /**
+   * Opens a repeatable prompt with a yes/no question and sets the output directory to the current user directory.
+   *
+   */
+  private void setOutputPathWithPrompt() {
```
maybeec, line 205: Cleaner solution to prevent others to program the next yes / no question would be to extract the actual logic from here. Put proper input and output parameters in place and make the question type reusable.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -140,14 +138,33 @@ public Integer doAction() throws Exception {
...
+            "Did not detect the input as part of a maven project, the root directory of the maven project was not found.");
+
+        // cancels with an error if more than one input file was provided
+        if (this.inputFiles.size() > 1) {
+          LOG.error(
+              "As multiple input files were provided, setting an automatic output path is not possible. Cancelling...");
+          System.exit(1);
+        }
+        LOG.info("Would you like to take '{}' as a root directory for output generation? \n"
+            + "type yes/y to continue or no/n to cancel (or hit return for yes).", System.getProperty("user.dir"));
```
maybeec, line 160: Why User dir?

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -140,14 +138,33 @@ public Integer doAction() throws Exception {
...
+
+        LOG.info(
+            "Did not detect the input as part of a maven project, the root directory of the maven project was not found.");
+
+        // cancels with an error if more than one input file was provided
+        if (this.inputFiles.size() > 1) {
+          LOG.error(
+              "As multiple input files were provided, setting an automatic output path is not possible. Cancelling...");
+          System.exit(1);
+        }
```
maybeec, line 158: I think we can remove this abortion for multiple files if we would simply propose to use pwd as the root output path

maybeec (changes requested):

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: user.dir = working directory (path where JVM was run from)
user.home = home directory
But if pwd is more reliable, I can change it.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: Added new UserAbortException and changed error code to 255.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: Moved the yes no prompt to another method.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: user.dir = working directory (path where JVM was run from)
user.home = home directory
But if pwd is more reliable, I can change it.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
jan-vcapgemini: Not sure about that, because if the user provides several input files, we can't ensure that he wants all files to be generated to the same directory.

maybeec: let's assume that, yes
Let's continue then, otherwise the user will be blocked all the time and he/she in any case has to provide one output path also by parameters. There is also no way of differentiation.

jan-vcapgemini: Ok, I can agree to that. Removed now.

maybeec (approved):

Issue #1430, opened 2021-11-19
Classpath cache files don't work with 2 different devon-ide environments


jan-vcapgemini:
### Expected behavior

As a CobiGen CLI user, I want CobiGen to cache my classpath for each devonfw IDE environment separately so that the classPathLoader uses my current environment dependencies properly.

### Actual behavior
When executing a generate command, a `pom-cp*.txt` file gets created in my `C:\Users\username\.cobigen\cli-config` folder.
When I execute another generate command in another environment, the same pom-cp file gets used and the classLoader tries to load classes from a different environment than my current.

```
[DEBUG] Current working directory: D:\devon_new\workspaces\main\testexample

...

[DEBUG] URL Classloader with URLs:
[DEBUG]   * file:/D:/devon_latest/conf/.m2/repository/com/devonfw/cobigen/javaplugin/2021.08.002/javaplugin-2021.08.002.jar
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Browse to your `C:\Users\username\.cobigen\cli-config` folder
2. Delete all `pom-cp*.txt` files
3. Debug the generate process in your IDE (look for: `[DEBUG] URL Classloader with URLs:`)
4. Create a 2nd devonfw-ide environment with a new workspace and project
5. Debug the generate process in your new IDE (again look for: `[DEBUG] URL Classloader with URLs:`)
6. The 2nd generate uses the classPaths of the first IDE environment

### Affected version:

* CobiGen CLI 2021.08.002

jan-vcapgemini:
Was temporarily fixed for new devonfw-ide installations with: https://github.com/devonfw/ide/issues/696.
Will be continued with: #1476

PR #1464, opened 2022-01-21
#1430 moved .cobigen directory path from user.home to maven .m2

jan-vcapgemini:
Adresses #1430.

@devonfw/cobigen

### DIFF
--- a/cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/CobiGenPaths.java
+++ b/cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/CobiGenPaths.java
@@ -37,7 +37,9 @@ public class CobiGenPaths {
           ConfigurationConstants.CONFIG_ENV_HOME, envValue);
       cobiGenPath = Paths.get(envValue);
     } else {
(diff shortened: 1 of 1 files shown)


jan-vcapgemini:
Couldn't the `COBIGEN_HOME` environment variable be added to devonfw-ide by default too?

e.g. `my-devon-ide/scripts/devon.properties`

`export COBIGEN_HOME=${DEVON_IDE_HOME}/conf/.m2/.cobigen`

Current solution repeatedly checks for the m2 repository folder using maven which is quite slow.

maybeec:
Mh you are putting now the cobigen home into the Maven repository. I think thats a bad idea. It's even not solving the problem as if the user is specifying any other directory, it will fail again.

maybeec:
> Couldn't the `COBIGEN_HOME` environment variable be added to devonfw-ide by default too?
> 
> e.g. `my-devon-ide/scripts/devon.properties`
> 
> `export COBIGEN_HOME=${DEVON_IDE_HOME}/conf/.m2/.cobigen`
> 
> Current solution repeatedly checks for the m2 repository folder using maven which is quite slow.
> 

Yes that could be done but the root cause of the problem would not be solved what about those who do not use the devonfw ide?

Issue #1462, opened 2022-01-20
Implement new reader, which is capable to restore the existing behavior by reading the new configuration structure


maybeec:

GuentherJulian:
The idea discussed in a first meeting was to remove the context.xml from the root folder and put the triggers directly in the subfolders. Either as a separate context.xml file in the subfolder or inside the templates.xml file.

PR #1467, opened 2022-01-26
#1462 adapted configuration reader to read new folder structure

GuentherJulian:
Adresses #1462

@devonfw/cobigen


cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationReader.java
```diff
@@ -216,4 +247,24 @@ public Path getContextRoot() {
...
+    if (Files.exists(contextPath)) {
+      this.contextFiles.add(contextPath);
+    } else {
+      File[] subFolders = directory.toFile().listFiles();
+      for (File subFolder : subFolders) {
+        if (subFolder.isDirectory()) {
+          addContextFilesRecursively(subFolder.toPath());
+        }
+      }
+    }
```
maybeec, line 268: this is quite some work and might cause stupid errors in case, someone copies a context.xml for backup to a separate place within the folders. Couldn't we be a little more restrictive on where we would expect context.xml files?

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationReader.java
```diff
@@ -59,19 +62,41 @@
...
+    this.contextFiles = new ArrayList<>();
+
+    Path contextFile = configRoot.resolve(ConfigurationConstants.CONTEXT_CONFIG_FILENAME);
+    if (!Files.exists(contextFile)) {
       configRoot = configRoot.resolve(ConfigurationConstants.TEMPLATE_RESOURCE_FOLDER);
-      this.contextFile = configRoot.resolve(ConfigurationConstants.CONTEXT_CONFIG_FILENAME);
-      if (!Files.exists(this.contextFile)) {
-        throw new InvalidConfigurationException(this.contextFile, "Could not find context configuration file.");
+      contextFile = configRoot.resolve(ConfigurationConstants.CONTEXT_CONFIG_FILENAME);
+      if (!Files.exists(contextFile)) {
```
maybeec, line 75: twice the same if?

maybeec (changes requested):

Issue #1461, opened 2022-01-20
Setup new templates folder per template set in https://github.com/devonfw/cobigen/tree/master/cobigen-templates and split `context.xml`


maybeec:

GuentherJulian:
The idea discussed in a first meeting was to remove the context.xml from the root folder and put the triggers directly in the subfolders. Either as a separate context.xml file in the subfolder or inside the templates.xml file.

LarsReinken:
CLI-tests use the currently deployed templates, therefore the current tests will fail the moment we deploy the new folder structure. the way CLI-tests work need to be changed, otherwise we can't create tests using the old folder structure and because of that can't test the backward compatibility.

MikeSchumacherCapgemini:
After more consideration, this is the folder structure we decided to use for now.

The Cobigen_Templates folder should contain any number of folders for template sets which contain a context and templates xml each as well as the templates in some way as would be denoted in the config files. The utils would stay completely seperate from the template sets and would be downloaded seperately (possibly prompting the user to download if they don't exist yet).

![Proof_Of_concept_Folder_Structure](https://user-images.githubusercontent.com/38068582/151963599-90d92444-7773-4d8e-b899-8a36809c7bc3.png)

The config files should always be top level in the TemplateSet folder, so situations like the following image would have to be avoided. This makes it easier to search for the config files (no searching in sub-folders required) and is kept in the spirit of modular templates, since you should only be able to download singular template sets.

![Bad_Folder_Structure](https://user-images.githubusercontent.com/38068582/151963602-0d1c486f-5548-4505-8c27-7f84a51702ca.png)

MikeSchumacherCapgemini:
In the future, the following might be a better solution that could be reached with multiple tweaks/extensions to the above solution. 

Just as in the previous comment, the Cobigen_Templates folder would contain any number of template sets, though there are no longer two config files on the top level, but one instead that contains the info of both files.
Additionally, not all utils would be downloaded at once, but instead each template set would have individual utils needed denoted in its config file.

![Optimal_Folder_Structure](https://user-images.githubusercontent.com/38068582/151965425-d5791ee9-22f1-44bc-988d-4237f3d4afe4.png)


Another proposed solution was to split up the util classes and use what each template sets needs in a new util file in each template set folder (no global utils).
This was disregarded for now, since there would be quite a bit of code duplication as many templates use the same util methods.

PR #1471, opened 2022-02-01
Adapt folder structure to new concept

MikeSchumacherCapgemini:
Adresses #1461 .

* Removed "constants" and "documentation" folders and pulled contained template sets to the top level of Cobigen_Templates
* Removed "templateFolder" attribute from "trigger" tag in context.xml as it is unnecessary (the attribute was previously used to find the folder the context file now resides in)

@devonfw/cobigen


cobigen-templates/templates-devon4j/src/main/templates/crud_angular_client_app/context.xml
```diff
@@ -2,18 +2,18 @@
 <contextConfiguration xmlns="http://capgemini.com/devonfw/cobigen/ContextConfiguration" version="2.1">
```
LarsReinken, line 2: since we need a new xsd, the version should be 2.2

LarsReinken (changes requested):
the version should be changed for every context.xml

LarsReinken (approved):

PR #1472, opened 2022-02-01
Added/adapted tests for new template folder structure

MikeSchumacherCapgemini:
Adresses #1462 

- Added test for context.xml not being top level
- Added test for context.xml in sub-folder
- Added test for both old context.xml and another modular one that checks for a thrown exception with a message that contains the relevant wiki link (relevant for old custom templates)
- Removed templateFolder attribute from context.xml files that are not in the root folder


cobigen/cobigen-core/src/test/java/com/devonfw/cobigen/unittest/config/reader/ContextConfigurationReaderTest.java
```diff
@@ -34,14 +34,51 @@ public void testErrorOnInvalidConfiguration() throws InvalidConfigurationExcepti
...
+  /**
+   * Tests whether a valid configuartion can be read from both root and templateSet folders (new templates with old
+   * custom templates) Backward Compatibility test, remove when monolithic context.xml is deprecated.
+   *
+   * @throws Exception test fails
+   */
+  @Test
+  public void testContextLoadedFromTemplateSetFolderAndRoot() throws Exception {
+
+    CobiGenFactory.create(new File(testFileRootPath + "valid_source_folder_modular_and_root").toURI());
```
GuentherJulian, line 81: Should we really consider this case?
Wouldn't it be better to support either the new or the old configuration and not a mixture of both?

MikeSchumacherCapgemini: Updated the test to expect a thrown exception with a message that contains the wiki link.

MikeSchumacherCapgemini:
Obsolete

PR #1474, opened 2022-02-07
Fix for GenerateCommandTest

GuentherJulian:
Implements

* Changed the increment in the GenerateCommandTest. The test failed because the increment number does not match with the new folder structure


cobigen-cli/cli-systemtest/src/test/java/com/devonfw/cobigen/cli/systemtest/GenerateCommandTest.java
```diff
@@ -290,7 +290,7 @@ public void generateFromArgsWithQuote() throws Exception {
     args[2] = "--out";
     args[3] = outputRootPath.toFile().getAbsolutePath();
     args[4] = "--increments";
-    args[5] = "15";
+    args[5] = "OpenAPI_Docs";
```
GuentherJulian, line 293: @LarsReinken The input is compared with the id (name) and the description of the increment. So here `OpenAPI_Docs` (id) and `Documentation` (description) both would work. See here https://github.com/devonfw/cobigen/blob/master/cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java#L538

Issue #1406, opened 2021-10-14
Generate with CRUD devon4j Server CRUD REST services fails with any input


jan-vcapgemini:
### Expected behavior

As a CobiGen Eclipse user, I want to generate a CRUD REST service artifact.

### Actual behavior
When generating from an input entity or openAPI file, following merge error occurs:

`Unable to merge the patch generated by template D:\devon_ide\workspaces\junit-workspace\CobiGen_Templates\src\main\templates\crud_openapi_java_server_app\templates\java\${variables.rootPackage}\${variables.component#lower_case}\service\api\rest\${variables.component#cap_first}RestService.java.ftl into file C:\Users\username\AppData\Local\Temp\cobigen-12921793936326071041\src\main\java\sampledatamanagement\service\api\rest\SampledatamanagementRestService.java: The syntax of the base file is invalid. Error in line: 1 / column: 9: syntax error @[1,9]
Unable to merge the patch generated by template D:\devon_ide\workspaces\junit-workspace\CobiGen_Templates\src\main\templates\crud_openapi_java_server_app\templates\java\${variables.rootPackage}\${variables.component#lower_case}\service\api\rest\${variables.component#cap_first}RestService.java.ftl into file C:\Users\username\AppData\Local\Temp\cobigen-12921793936326071041\src\main\java\sampledatamanagement\service\api\rest\SampledatamanagementRestService.java: The syntax of the base file is invalid. Error in line: 1 / column: 9: syntax error @[1,9]
  com.devonfw.cobigen.api.exception.MergeException: Unable to merge the patch generated by template D:\devon_ide\workspaces\junit-workspace\CobiGen_Templates\src\main\templates\crud_openapi_java_server_app\templates\java\${variables.rootPackage}\${variables.component#lower_case}\service\api\rest\${variables.component#cap_first}RestService.java.ftl into file C:\Users\username\AppData\Local\Temp\cobigen-12921793936326071041\src\main\java\sampledatamanagement\service\api\rest\SampledatamanagementRestService.java: The syntax of the base file is invalid. Error in line: 1 / column: 9: syntax error @[1,9]
  	at com.devonfw.cobigen.impl.generator.GenerationProcessorImpl.generate(GenerationProcessorImpl.java:525)
  	at com.devonfw.cobigen.impl.generator.GenerationProcessorImpl.generate(GenerationProcessorImpl.java:212)
[...]

PR #1478, opened 2022-02-08
Update config folder structure and reader for modular template sets

MikeSchumacherCapgemini:
Implements

* New folder Structure for modular template sets
* New configuration reader tests for the above
* Changes to the configuration reader

Fixes

* #1430
* #1406

@devonfw/cobigen


cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/upgrade/ContextConfigurationUpgrader.java
```diff
@@ -37,24 +37,24 @@ protected ConfigurationUpgradeResult performNextUpgradeStep(ContextConfiguration
 
     switch (source) {
       case v2_0:
-        // to v2.1
+        // to v2.2
```
maybeec, line 40: you are missing the case for v2_1, which will be the same

maybeec (changes requested):

maybeec (changes requested):

maybeec:
please check for the last comments and finalize the PR and give it for review officially if you think it's ready and doesn't break backward compatibility so we can merge it to master

maybeec (approved):

maybeec (approved):

PR #1504, opened 2022-03-21
1452 Implemented new configuration reader to read the template sets

### LINKED ISSUE #1452: Split template sets in separate deployables
As a cobigen developer, I want to deploy every template set in a separate deployable so that I can release each template set separately. Please setup all template sets under the groupId `com.devonfw.cobigen.templates`.

* [x] #1501
* [x] #1505
* [x] #1504
* [x] adapt templates should extract all installed sets (short term solution for #1509 / #1455)

GuentherJulian:
Adresses/Fixes #1452 /  #1505

Implements

*  Implemented new `ContextConfigurationSetReader` for the new template set folder structure. The reader prefers the templates in the adapted folder rather than the template jar files in the downloaded folder
* Add new tests to `ContextConfigurationReaderTest` to test the new structure
* Changed CLI and template tests to use a copy of the current template project in their tests
* Adapted the implementation for loading the util classes so that the utils now can be loaded from the `devon4j-utils-project`. Either direct from the jar or from the extracted jar.

@devonfw/cobigen


cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
```diff
@@ -1,21 +1,170 @@
...
+
+  @Override
+  public Map<String, Trigger> loadTriggers() {
+
+    Map<String, Trigger> triggers = Maps.newHashMap();
+    for (Path contextFile : this.contextConfigurations.keySet()) {
+      Path configRoot = this.configRoots.get(contextFile);
+      ContextConfiguration contextConfiguration = this.contextConfigurations.get(contextFile);
+      boolean isJarConfig = (configRoot.getParent() == null);
+      for (com.devonfw.cobigen.impl.config.entity.io.Trigger t : contextConfiguration.getTrigger()) {
```
LarsReinken, line 142: with the new configuration just one trigger is allowed. maybe we should should use just the first to be compliant with the get Trigger method but make sure we don't have any context.xml files which use more than one trigger.

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
```diff
@@ -1,21 +1,170 @@
...
+        String templateFolder;
+        if (isJarConfig) {
+          templateFolder = contextFile.getParent().toString();
+        } else {
+          templateFolder = configRoot.getFileName().resolve(ConfigurationConstants.TEMPLATE_RESOURCE_FOLDER).toString();
+          configRoot = configRoot.getParent();
```
LarsReinken, line 148: this would result in problems, when we have more than one trigger, because for each trigger the configRoot would set to the parent folder.

LarsReinken (changes requested):

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationFinder.java
```diff
@@ -130,18 +131,24 @@ private static URI findTemplates(Path home) {
       return templatesFolderPath.toUri();
     }
 
-    // 2 create/use new template sets folder
+    // 2. use template jar
+    File jarFile = TemplatesJarUtil.getJarFile(false, templatesFolderPath.getParent().toFile());
+    if (jarFile != null) {
+      return jarFile.toPath().toUri();
+    }
```
jan-vcapgemini, line 138: Please use Path instead of File.
For the if condition you can use Files.exists(jarFile)

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
```diff
@@ -1,21 +1,170 @@
...
+          triggers.put(t.getId(), new Trigger(t.getId(), t.getType(), templateFolder,
+              Charset.forName(t.getInputCharset()), loadMatchers(t), loadContainerMatchers(t)));
+        }
+      }
+    }
+    return triggers;
+  }
+
+  /**
+   * @param triggerId the trigger id to search the config root for
```
jan-vcapgemini, line 163: Missing description.

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
```diff
@@ -1,21 +1,170 @@
...
+
+    try (Stream<Path> files = Files.list(configRoot)) {
+      files.forEach(path -> {
+        if (path.toUri().toString().endsWith(".jar")) {
+          templateJars.add(path);
+        }
+      });
+    } catch (IOException e) {
+      throw new InvalidConfigurationException(configRoot, "Could not read configuration root directory.", e);
+    }
```
jan-vcapgemini, line 118: Can be replaced with TemplatesJarUtil.getJarFiles.

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
```diff
@@ -1,21 +1,170 @@
...
+    for (Path jarPath : templateJars) {
+      Path configurationPath = FileSystemUtil.createFileSystemDependentPath(jarPath.toUri());
+      Path contextPath = configurationPath.resolve(ConfigurationConstants.TEMPLATE_RESOURCE_FOLDER)
+          .resolve(ConfigurationConstants.CONTEXT_CONFIG_FILENAME);
+
+      if (Files.exists(contextPath)) {
+        contextPaths.add(contextPath);
+        this.configRoots.put(contextPath, configurationPath);
+      }
+    }
```
jan-vcapgemini, line 129: Should be moved to another method.

jan-vcapgemini (changes requested):

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
GuentherJulian: Changed it so that only the first entry is used. Do you mean that also an error should be thrown if there is more than one trigger in a context config file?

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationFinder.java
GuentherJulian: This would cause a NullPointerException, since getJarFile returns null if nothing was found.
We could change the return value of getJarFile, but then we also need to adjust some other classes that use the method.

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/reader/ContextConfigurationSetReader.java
LarsReinken: Because of the assumption, that just one trigger in a contextFile is possible that point isn't relevant anymore.

LarsReinken (approved):

cobigen-cli/cli-systemtest/src/test/java/com/devonfw/cobigen/cli/systemtest/AbstractCliTest.java
```diff
@@ -32,23 +35,78 @@
...
+      try (Stream<Path> files = Files.list(templateSetsAdaptedFolder)) {
+        files.forEach(path -> {
+          if (Files.isDirectory(path)) {
+            Path resourcesFolder = path.resolve("src/main/resources");
+            Path templatesFolder = path.resolve(ConfigurationConstants.TEMPLATE_RESOURCE_FOLDER);
+            if (Files.exists(resourcesFolder) && !Files.exists(templatesFolder)) {
+              try {
+                Files.move(resourcesFolder, templatesFolder);
+              } catch (IOException e) {
+                e.printStackTrace();
```
jan-vcapgemini, line 89: Please throw an exception with a message.

cobigen-cli/cli-systemtest/src/test/java/com/devonfw/cobigen/cli/systemtest/AbstractCliTest.java
```diff
@@ -32,23 +35,78 @@
...
+              }
+            }
+          }
+          if (path.getFileName().toString().equals("templates-devon4j-utils")) {
+            if (Files.exists(path.resolve("pom.xml"))) {
+              try {
+                Files.delete(path.resolve("pom.xml"));
+              } catch (IOException e) {
+                e.printStackTrace();
```
jan-vcapgemini, line 98: Please throw an exception with a message.

cobigen-cli/cli-systemtest/src/test/java/com/devonfw/cobigen/cli/systemtest/AbstractCliTest.java
```diff
@@ -32,23 +35,78 @@
...
+              }
+            }
+            try {
+              Files.copy(utilsPom, path.resolve("pom.xml"));
+            } catch (IOException e) {
+              e.printStackTrace();
```
jan-vcapgemini, line 104: Please throw an exception with a message.

jan-vcapgemini (changes requested):

jan-vcapgemini (approved):

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/util/ConfigurationFinder.java
jan-vcapgemini: Not if you do it like jarFile != null && Files.exists(jarFile).
Btw. shall we keep the deprecated annotation of getjarFile?

GuentherJulian: Not sure if I get your intention. `Files.exists` works only on `Path` not on `File`. So again we had to change the return type of `getJarFile`, which I wanted to avoid.

> Btw. shall we keep the deprecated annotation of getjarFile?
I am not sure if this would be correct right now, as the old structure is still supported, so we may need to continue using `getJarFile` somewhere else

PR #1535, opened 2022-05-12
1511 enable custom template set group ids for lookup

### LINKED ISSUE #1511: Enable custom template-set groupIds for lookup
As an architect, I want to have control over the sources of the template-sets so that I can provide my custom template-sets to development and even restrict the teams to ask for default templates provided by cobigen.

*Acceptance Criteria*
* [x] I want to be able to specify further maven groupIds, where cobigen will search for template-set artifacts by the configuration key `template-sets.groupIds` in the cobigen configuration
* [x] I want to be able to configure multiple (comma separated) groupIds 
* [x] By default (public) cobigen groupId will anyhow be continued to be queried to allow public template-set lookup simultaneously
* [x] By the configuration key `template-sets.allow-snapshots=true`, I want to be able to allow snapshots of template-sets to be offered for template-set development purposes. By default no snapshots should be queried.
* [x] By the configuration key `template-sets.disable-default-lookup=false`, I want to be able to disable by default querying of default public groupIds preconfigured in cobigen.
* [x] By the configuration key `template-sets.hide`, I want to be able to hide very specific template sets or versions of template sets. Thus, I want to be able to specify a comma separated list of maven coordinates of the following pattern `groupId(:artifactId(:version)?)?`
* [x] All configuration is documented in the wiki and finally also occurs on the website

mdukhan:
Adresses/Fixes #1511 

@devonfw/cobigen


EduardKrieger (approved):

documentation/cobigen-core_configuration.asciidoc
```diff
@@ -354,7 +354,7 @@ configure multiple (comma separated) `groupIds`. By default, (public) CobiGen `g
 template-sets.groupIds=com.devonfw.cobigen.templates,jaxen,jakarta.xml.bind
 template-sets.allow-snapshots=true
 template-sets.disable-default-lookup=false
-template-sets.hide=com.devonfw(:test-artifact(:3.2.1-SNAPSHOT))
+template-sets.hide=com.devonfw.cobigen:core-parent:2021.12.007-SNAPSHOT
```
GuentherJulian, line 357: I would use a different example here for the documentation since `com.devonfw.cobigen:core-parent` is a valid expression but makes no sense (cobigen core does not contain template-sets). Use either a hypothetical example or one of the standard CobiGen template sets.

GuentherJulian (changes requested):

documentation/cobigen-core_configuration.asciidoc
mdukhan: com.devonfw.cobigen.templates:templates-parent:2021.12.006-SNAPSHOT
com.devonfw.cobigen.templates:crud-angular-client-app:2021.12.007-SNAPSHOT

These will be valid examples correct ?

GuentherJulian (approved):

maybeec (approved):

Issue #1531, opened 2022-05-09
Inform the user about deprecation and upgrade options


maybeec:
As a cobigen user, I want to get automatically informed about the conceptual changes of the configuration going from monolithic template configuration folder to a modular template-set configuration folder. I want to get offered information (#1516).

*Acceptance Criteria*
* [x] If cobigen reads my monolithic template configuration, which is deprecated, I want to get offered an upgrade option and get informed about the process and steps to be performed
* [x] To properly make my decision in the moment of time, I would like to get more information about why I should do the upgrade and what is the impact 
* [x] the user should be asked every time a monolithic configuration is read from a folder 
* [x] make sure, that maven execution is not blocked by this feature at all

maybeec:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/CobiGen-StoryMap-6257b8b9ecf3920015f4e871/issues/devonfw/cobigen/1531?planning-poker) with ZenHub @EduardKrieger @GuentherJulian @jan-vcapgemini @LarsReinken @mdukhan

mdukhan:
In Ticket #1519 we had to Ignore all Eclipse tests and some CLI tests, In this ticket these Tests should be enabled, these are: 

* All Eclipse Tests
* All tests in GenerateCommandTest in CLI

PR #1538, opened 2022-05-25
#1519 added possible Log warning places for CLI

### LINKED ISSUE #1519: Add disclaimer of deprecation of monolithic configuration

mdukhan:
Adresses/Fixes #1519 .

@devonfw/cobigen

### DIFF
--- a/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/GeneratorWrapperFactory.java
+++ b/cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/GeneratorWrapperFactory.java
@@ -236,7 +236,7 @@ public class GeneratorWrapperFactory {
     } catch (Throwable e) {
       throw new GeneratorCreationException(
           "Configuration source could not be read.\nIf you were updating templates, it may mean"
-              + " that you have no internet connection.",
+              + " that you have no internet connection," + " or you have an old monolithic Templates.",
           e);
     }
   }
--- a/cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/exception/DeprecatedMonolithicTemplatesException.java
+++ b/cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/exception/DeprecatedMonolithicTemplatesException.java
@@ -1,0 +1,35 @@
+package com.devonfw.cobigen.api.exception;
+
+/**
+ * Exception that indicates that an old monolithic template structure has been found. For asking if the template
+ * structure should be upgraded later.
+ */
+public class DeprecatedMonolithicTemplatesException extends CobiGenRuntimeException {
+
+  /**
+   * Default serial version UID.
+   */
+  private static final long serialVersionUID = 1;
+
+  /**
+   * The constructor.
+   *
+   * @param message informative message describing the exception or where it has been occurred
+   * @param cause originating this exception
+   */
+  public DeprecatedMonolithicTemplatesException(String message, Throwable cause) {
+
+    super(message, cause);
+  }
+
+  /**
+   * The constructor.
+   *
+   * Creates a new {@link DeprecatedMonolithicTemplatesException} with a proper notification message
+   */
+  public DeprecatedMonolithicTemplatesException() {
+
+    super("You Are using an old Templates Configuration, please consider to upgrade your templates! Thank You!");
+  }
+
+}
--- a/cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
+++ b/cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
@@ -119,7 +119,6 @@ public class ConfigurationHolder {
 
     if (this.configurationPath.toUri().getScheme().equals("jar")
         || !this.configurationPath.getFileName().toString().equals(ConfigurationConstants.TEMPLATE_SETS_FOLDER)) {
-      LOG.warn("You are using an old templates version, please make sure to upgrade to the newest three!");
       return false;
     }
     return true;
--- a/cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
+++ b/cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
@@ -8,10 +8,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.devonfw.cobigen.api.constants.ConfigurationConstants;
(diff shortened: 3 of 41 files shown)


LarsReinken (changes requested):
I don't think a log is enough to inform the user. We should rather build a solution that works for both eclipse and CLI and in the case of CLI requires a user input to continue with the old configuration.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/GeneratorWrapperFactory.java
```diff
@@ -233,11 +241,23 @@ private static CobiGen initializeGenerator() throws InvalidConfigurationExceptio
       }
     } catch (CoreException e) {
       throw new GeneratorCreationException("An eclipse internal exception occurred", e);
-    } catch (Throwable e) {
+    } catch (GeneratorCreationException e) {
```
maybeec, line 244: Be careful, you now miss any other exception to be catched here.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/GeneratorWrapperFactory.java
```diff
@@ -61,9 +62,12 @@ public class GeneratorWrapperFactory {
    * @throws GeneratorProjectNotExistentException if the generator configuration project does not exist
    * @throws InvalidInputException if the selection includes non supported input types or is composed in a non supported
    *         combination of inputs.
+   * @throws UpgradeTemplatesNotificationException
+   * @throws InvalidConfigurationException
    */
   public static CobiGenWrapper createGenerator(ISelection selection, IProgressMonitor monitor)
-      throws GeneratorCreationException, GeneratorProjectNotExistentException, InvalidInputException {
+      throws GeneratorCreationException, GeneratorProjectNotExistentException, InvalidInputException,
+      InvalidConfigurationException, UpgradeTemplatesNotificationException {
```
maybeec, line 70: 1. Please rename the exception. What is the exception about? It's not an exception because of "upgrade template notification". That's actually already encoding how you want to handle it. But the exception name should be the cause and not imply any handling as this can differ from case to case.
2. Please make UpgradeTemplatesNotificationException as unchecked exception and remove the throws from all method declaration. As you might have noticed, it's hell of work to continuously declare all exceptions in every method. So normally we should not declare the exception at every method, but handle them at the points of interest. Unchecked exceptions should be preferred.

cobigen-cli/cli/src/main/java/com/devonfw/cobigen/cli/commands/GenerateCommand.java
```diff
@@ -125,8 +129,9 @@ public Integer doAction() throws Exception {
    * Resolves dependencies from templates
    *
    * @throws IOException
+   * @throws UpgradeTemplatesNotificationException
    */
-  private void resolveTemplateDependencies() throws IOException {
+  private void resolveTemplateDependencies() throws IOException, UpgradeTemplatesNotificationException {
```
maybeec, line 134: I am not sure, why I cant find the exception  UpgradeTemplatesNotificationException as part of this PR. Has it been integrated beforehand in another PR? 
Is this Exception inheriting the Cobigen creation exception?

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/healthcheck/AdvancedHealthCheckDialog.java
```diff
@@ -141,7 +142,12 @@ public void widgetSelected(SelectionEvent e) {
               MDC.put(InfrastructureConstants.CORRELATION_ID, UUID.randomUUID().toString());
               AdvancedHealthCheckDialog.this.healthCheck.upgradeTemplatesConfiguration(
                   AdvancedHealthCheckDialog.this.upgradeableConfigurations.get(key), BackupPolicy.BACKUP_IF_POSSIBLE);
-              refreshUI();
+              try {
+                refreshUI();
+              } catch (UpgradeTemplatesNotificationException e1) {
+                // create New UI Dialog for healthcheck
+                PlatformUIUtil.openErrorDialog("You are using an old Templates version!", e1);
+              }
```
maybeec, line 150: I have seen different catch blocks with redundant exception handling, Please make sure, that this is properly and consistently handled. I think there is even an ExceptionHandler class insider of the eclipse plugin, which you could use to handle the exception centrally and consistently sharing code.

cobigen-eclipse/cobigen-eclipse/src/com/devonfw/cobigen/eclipse/generator/GeneratorWrapperFactory.java
```diff
@@ -233,11 +241,23 @@ private static CobiGen initializeGenerator() throws InvalidConfigurationExceptio
...
     }
   }
+
+  private static Path createWidgetFromException() {
+
+    try {
+      return CobiGenPaths.getTemplatesFolderPath();
+    } catch (Exception e) {
+      // TODO Auto-generated catch block
+      // Create UI WARN MESSAGE TO NOTIFY USER FOR UPGRADE
```
maybeec, line 258: please remove the comments, they don't give any value.

maybeec (changes requested):

maybeec (approved):

LarsReinken (approved):

LarsReinken (changes requested):
I don't think an exception works to transfer the information to CLI and Eclipse because it blocks subsequent generation.

LarsReinken (dismissed):

cobigen/cobigen-core/src/main/java/com/devonfw/cobigen/impl/config/ConfigurationHolder.java
```diff
@@ -20,6 +24,9 @@
  */
 public class ConfigurationHolder {
 
+  /** Logger instance */
+  private static final Logger LOG = LoggerFactory.getLogger(ConfigurationFinder.class);
```
GuentherJulian, line 28: Use `ConfigurationHolder.class` instead

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/exception/DeprecatedMonolithicTemplatesException.java
```diff
@@ -0,0 +1,35 @@
+package com.devonfw.cobigen.api.exception;
+
+/**
+ * Exception that indicates that an old monolithic template structure has been found. For asking if the template
+ * structure should be upgraded later.
+ */
+public class DeprecatedMonolithicTemplatesException extends CobiGenRuntimeException {
```
GuentherJulian, line 7: Where is this Exception catched and handled?

LarsReinken: it's part of ticket #1531 to handle the exception

PR #1550, opened 2022-06-29
Merge of master into template-set-deployables

jan-vcapgemini:
Just the merge of the master.
Please double check because I've got way too many merge conflicts.

@devonfw/cobigen


jan-vcapgemini:
The Eclipse tests are failing because the generateWithTextInputDoesNotLoop test is using an old templates configuration. Normally we set these tests to ignore till the upgrade/postpone process is implemented.
I could upgrade the resource test template to 3.0 if you think this is required.

Issue #1548, opened 2022-06-28
devon4j templates: SearchCriteriaTo generated in wrong package.


hohwille:
### Expected behavior

As a devonfw user, I want to use Cobigen with devon4j and generate compliant code so that sonar-devon4j-plugin accepts the generated code without issues.

### Actual behavior

The generated code violates the devon4j architecture since `SearchCriteriaTo` objects are generated in `logic` layer while they are used in DAO/repository located in `dataaccess` layer. This is resulting in a code dependency from `dataaccess` layer to `logic` layer what is an access upwards the layers what is forbidden by a layered architecture (see https://github.com/devonfw/sonar-devon4j-plugin/blob/master/documentation/rules.asciidoc and  https://github.com/devonfw/devon4j/blob/master/documentation/architecture.asciidoc#technical-architecture or (German) Wikipedia).

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Generate Code from Entity including SearchCriteriaTo and DAO/Repo
2. Analyze the code with sonar-devon4j-plugin
3. See the findings (https://github.com/devonfw/sonar-devon4j-plugin/issues/138)

### Related/Dependent Issues

https://github.com/devonfw/sonar-devon4j-plugin/issues/138

### Comments/Hints:

https://github.com/devonfw/cobigen/blob/6cd6ebb096d97b5ce539383aa16d8d97cbf7b92b/cobigen-templates/templates-devon4j/src/main/templates/crud_java_server_app/templates/java/%24%7Bvariables.rootPackage%7D/%24%7Bvariables.component%7D/logic/api/to/%24%7Bvariables.entityName%7DSearchCriteriaTo.java.ftl#L2
https://github.com/devonfw/cobigen/blob/6cd6ebb096d97b5ce539383aa16d8d97cbf7b92b/cobigen-templates/templates-devon4j/src/main/templates/crud_java_server_app/templates/java/%24%7Bvariables.rootPackage%7D/%24%7Bvariables.component%7D/dataaccess/api/repo/%24%7Bvariables.entityName%7DRepository.java.ftl#L19

Simple solution is to move the package-segment of the SearchCriteriaTo from `logic` to `common`.

### Affected version:

* 2021.12.006

maybeec:
please move the search criteria to commons package

PR #1560, opened 2022-08-01
#1548 SearchCriteriaTo generated in wrong package

ZakariaeMkassi:
The package where the SearchCriteriaTo object is generated is wrong, so the template has been moved from logic to common package, and all places where the object is imported are modified.


EduardKrieger (approved):

EduardKrieger:
@ZakariaeMkassi Please sign the CLA

PR #1610, opened 2022-11-25
1595 extend uprader to upgrade from monolithic structure to template set.xml

### LINKED ISSUE #1595: Extend the upgrader to upgrade from monolithic structure to template-set.xml
As a Cobigen user, I want to be able to let Cobigen upgrade my adapted template configuration to the new merged template-set.xml (#1495 ).

*Acceptance Criteria*
* [x] the upgrader from #1518  should be used and extended accordingly. 
* [x] upgrade users context.xml, including mergeing it with templates.xml into template-set.xml
* [x] After upgrading to the new configuration, the configuration should be re-read and also used #1522

quantumfate:
Adresses/Fixes #1595 .

Implements

* merge from context.xml and template.xml
* context.xml will be deleted and adapted to new template-set.xml

@devonfw/cobigen

### DIFF
--- a/cobigen-cli/cli/pom.xml
+++ b/cobigen-cli/cli/pom.xml
@@ -31,7 +31,7 @@
     <dependency>
       <groupId>info.picocli</groupId>
       <artifactId>picocli</artifactId>
-      <version>4.6.2</version>
+      <version>4.7.0</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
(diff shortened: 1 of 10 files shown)


Lur1an:
The current implementation (on this branch) extends the upgrade function from V2.1 to V3 and directly generates a template-set.xml (V6) and skips the new context.xml generation. In doing so it breaks a lot of core code that expects new context.xml files from the upgrade.

I suggest to integrate this upgrader by skipping the V3 upgrade altogether and go directly from V2.1 Monolithic context.xml to template-sets.xml V6. This will require reworking the AbstractConfigurationUpgrader and related classes/methods.

Hint:
- check failing tests in core to see where to start reworking.

PR #1613, opened 2022-12-01
Optimize class path cache files

### LINKED ISSUE #1476: Optimize class path cache files handling/generation
Continuation of: #1430

As a CobiGen CLI user, I want CobiGen to handle my class path cache files more effective and less error prone.

- [x] Conceptualize new class path cache file handling/generation
- [x] Validate class path cache file integrity (check if dependencies exist)

EduardKrieger:
Adresses/Fixes #1476.
Initial version to cache the dependencies more efficient and less error prone.

Implements

* generatePomFileHash reworked also takes m2 repo into account.
* validation of the existing cache, checking if file exists and is in the correct repository
* added unit tests for both functionalities 
* Integration test with the new cache functions

@devonfw/cobigen


cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -154,24 +163,66 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+        throw e;
+      }
+    }
+  }
+
   /**
    * Generates a hash for the provided POM file
    *
    * @param pomFile to generate hash from
+   * @param m2RepoPath TODO
```
maybeec, line 215: open todo

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -154,24 +163,66 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+          throw new IllegalPathStateException();
+        }
+        File cpFile = new File(cp);
+        if (!cpFile.exists()) {
+          LOG.warn("Cache {} is outdated, the dependencies will be resolved again", cp.toString());
+          throw new FileNotFoundException();
+        }
+      } catch (URISyntaxException e) {
+        LOG.error("Error while reading files from Cache");
+        throw e;
```
maybeec, line 206: log or throw. Also just throw proper exceptions with a message and cobigen runtime type

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -154,24 +163,66 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+        if (!Paths.get(cp).startsWith(Paths.get(repo))) {
+          LOG.warn(
+              "Cache {} pointing to another maven Repository, this could cause some problems, the dependencies will be resolved again",
+              cp.toString());
+          throw new IllegalPathStateException();
+        }
+        File cpFile = new File(cp);
+        if (!cpFile.exists()) {
+          LOG.warn("Cache {} is outdated, the dependencies will be resolved again", cp.toString());
+          throw new FileNotFoundException();
```
maybeec, line 202: log or throw. Also just throw proper exceptions with a message and cobigen runtime type

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -154,24 +163,66 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+
+    URI repo = determineMavenRepositoryPath().toUri().toURL().toURI(); // to change /// to / in the URI
+    for (URL classPath : classPathEntries) {
+      try {
+        URI cp = classPath.toURI();
+        if (!Paths.get(cp).startsWith(Paths.get(repo))) {
+          LOG.warn(
+              "Cache {} pointing to another maven Repository, this could cause some problems, the dependencies will be resolved again",
+              cp.toString());
+          throw new IllegalPathStateException();
```
maybeec, line 197: log or throw. Also just throw proper exceptions with a message and cobigen runtime type

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -39,6 +43,9 @@ public class MavenUtil {
   /** Logger instance. */
   private static final Logger LOG = LoggerFactory.getLogger(MavenUtil.class);
 
+  // maven repository
+  private static Path MAVEN_REPOSITORY = null;
```
maybeec, line 47: Needs to be properly tested in all environments whether the maven repository might not be incorrectly cached. The main question would be whether there are valid situations, where the class might be loaded and thus the maven repository might be cached, but the maven repository might then be changed intermediately.
Second question might be, what might happen if the repository is incorrectly cached? Will that break anything?

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -39,6 +43,9 @@ public class MavenUtil {
   /** Logger instance. */
   private static final Logger LOG = LoggerFactory.getLogger(MavenUtil.class);
 
+  // maven repository
```
maybeec, line 46: define proper javadoc

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -154,24 +163,66 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
             }
             return null;
           }).toArray(size -> new URL[size]);
-
+      try {
+        validateCachedClassPaths(classPathEntries);
+      } catch (FileNotFoundException | URISyntaxException | IllegalPathStateException e) {
```
maybeec, line 168: actually this is strange to throw and catch so many exceptions which actually do the same

maybeec (changes requested):

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
EduardKrieger: Added description

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
EduardKrieger: Removed throws, function now returns a boolean

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
EduardKrieger: Removed throws, function now returns a boolean

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
EduardKrieger: Removed throws, function now returns a boolean

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
EduardKrieger: I started to look into this problem with @jan-vcapgemini and will collect our thoughts here:

- **Needs to be properly tested in all environments whether the maven repository might not be incorrectly cached.**
  - What do you mean by incorrectly cached repository? The only possibility we could think about is that while cobigen is running (just in the generate process) the local repository setting will be changed by the user. This would result into the usage of the old incorrect repository and could not be detected. 
  - The detection of the local repository in MavenUtil depends on the maven executable path that already will be cached in SystemUtil.
  - We added an existence check to the cache that will be revaluated if the path doesn't exists. That's the only way we could think of checking for invalid cache.
  - We think the usage of CLI cobigen and the maven plugin could not produce problems, but we are unsure how we could invalidate the cache inside the eclipse cobigen plugin at runtime. I have tried to change the class path variable maven_home within the class path variables settings and also changing the repository in the maven settings but could not break cobigen.
- **The main question would be whether there are valid situations, where the class might be loaded and thus the maven repository might be cached, but the maven repository might then be changed intermediately.**
  -  We couldn't think of a usecase for every runtime where this might occur.
- **Second question might be, what might happen if the repository is incorrectly cached? Will that break anything?**
  - An incorrect Path that points to and existing repository by editing the settings.xml from the user. This situation could not be detected and already exists in our current state of the code by caching the maven executable in SystemUtil. Also this would result into loading some dependencies from another repository, or they would be downloaded again.
  - An incorrectly cached repository that does not exist could break the generation, but a check will be added.

EduardKrieger: As we discussed, I'm creating a new issue out of this problem, since it exceeds the scope of the current issue. There I will summarize the conversation and findings so far.

EduardKrieger: This issue will be continued in #1617.

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -155,23 +159,63 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+              "Cache {} pointing to another maven Repository, this could cause some problems, the dependencies will be resolved again",
+              cp.toString());
+          return false;
+        }
+        if (!Files.exists(cp)) {
+          LOG.warn("Cache {} is outdated, the dependencies will be resolved again", cp.toString());
+          return false;
+        }
+      } catch (URISyntaxException e) {
+        LOG.warn("Error while reading files from Cache, cache should ");
```
mdukhan, line 197: missing sentence "cache should ..."

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -155,23 +159,63 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
   /**
-   * Generates a hash for the provided POM file
+   * Validating the cached classpath entries and resolving missing files and dependencies from a repository that is not
+   * the current maven repository with a update of the classpath cache
+   *
+   * @param classPathEntries URLs of the cached classPath entries
+   * @throws URISyntaxException
+   * @throws FileNotFoundException
+   * @throws MalformedURLException
+   */
```
mdukhan, line 179: Maybe a small comment when does the method return true or false?

mdukhan (changes requested):

mdukhan (approved):

Ali7223 (dismissed):

cobigen/cobigen-core-api/src/main/java/com/devonfw/cobigen/api/util/MavenUtil.java
```diff
@@ -155,23 +158,62 @@ public static URLClassLoader addURLsFromCachedClassPathsFile(Path classPathCache
...
+      }
       return new URLClassLoader(classPathEntries, parentClassLoader);
     } catch (IOException e) {
       throw new CobiGenRuntimeException("Unable to read " + classPathCacheFile, e);
     }
   }
 
   /**
-   * Generates a hash for the provided POM file
+   * Validating the cached classpath entries and resolving missing files and dependencies from a repository that is not
```
jan-vcapgemini, line 171: Better: Validates the ... and resolves missing... with an update of...

jan-vcapgemini (changes requested):

jan-vcapgemini (approved):

maybeec (approved):

PR #1621, opened 2023-01-03
Dotnet templates webapi

JosepFe:
This pull request includes the new version of the templates for devon4net

Implements

New Cobigen templates for devon4net

@devonfw/cobigen


JosepFe:
The secret that the GitGuardian is detecting is a previously existing secret in the repository, this secret is local and for test uses only. I can't remove this secret from previous commits as I don't have permissions on the repository and it must be done by an admin. 

With the new changes, this secret is completely removed.

PR #1632, opened 2023-01-27
#158: MVP level 0 for language agnostic templates

### LINKED ISSUE #158: language agnostic templates

hohwille:
Adresses/Fixes #158

Implements

* a new template engine as cobigen-plugin
* allows to resolve variables as described in #158
* also strips out all lines containing the keyword "CobiGen" (case-sensitive). This is planned to allow constructs in your code like @CobenGenTemplate, @CobiGenCondition, CobiGenMacroXyz.run() etc. that should not appear in the output including according import statements.

@devonfw/cobigen


hohwille:
The broken build does not seem to be related to my changes but rather a GHA maintenance issue:

>  Error: This version of the CodeQL Action was deprecated on January 18th, 2023, and is no longer updated or supported. For better performance, improved security, and new features, upgrade to v2. For more information, see https://github.blog/changelog/2023-01-18-code-scanning-codeql-action-v1-is-now-deprecated/

cobigen-plugins/cobigen-templateengines/cobigen-tempeng-agnostic/src/main/java/com/devonfw/cobigen/tempeng/agnostic/AgnosticTemplateEngine.java
```diff
@@ -0,0 +1,127 @@
...
+ * choice can be used for auto-completion, refactoring, code-formatting, etc. In order to represent variables in the
+ * template, you simply put them in a specific syntax that is compliant with any
+ */
+@Name("Agnostic")
+public class AgnosticTemplateEngine implements TextTemplateEngine {
+
+  /** Logger instance. */
+  private static final Logger LOG = LoggerFactory.getLogger(AgnosticTemplateEngine.class);
+
+  private static final Pattern PATTERN_VARIABLE = Pattern.compile("(\\.?)([xX]_([a-zA-Z][a-zA-Z0-9_$-]*)_[xX])");
```
hohwille, line 37: There seems to be some "duplication" between `Variables` from core and this class.
We should discuss and align this. IMHO it is good to have this replacement logic centrally inside core and have only a single REGEX for this agnostic variable syntax.
For design we should think if `Variables` is really the right place for this. IMHO we should extract a dedicated and reusable class for only the replacement and variable resolution logic and use it from here and from `Variables`. WDYT?

hohwille:
This PR is more or less replaced with PR #1634 (level 1).

hohwille:
Closing as replaced with PR #1636

PR #1646, opened 2023-03-20
#158: Language-agnostic-templates merged for template-set-deployables

### LINKED ISSUE #158: language agnostic templates

hohwille:
Adresses/Fixes #158

See PR #1636 that was created for `master` branch and gets no attention/feedback.

@devonfw/cobigen


hohwille:
Build faile with:
```
Error:  Failed to execute goal on project tempeng-agnostic: Could not resolve dependencies for project com.devonfw.cobigen:tempeng-agnostic:jar:2021.12.007-SNAPSHOT: Could not find artifact com.devonfw.cobigen.templates:templates-devonfw-java:jar:2021.12.007-SNAPSHOT in ossrh (https://s01.oss.sonatype.org/content/repositories/snapshots/) -> [Help 1]
```
IMHO the dependency is there but the build setup is kind of strange as it does not build the entire maven reactor with all modules. So to me it looks rather like a problem of the build process and actions with its caching magic rather than a real problem but I might miss something...

hohwille:
This is my local build:
```
...
[INFO] CobiGen - Templates Parent ......................... SUCCESS [  0.358 s]
[INFO] CobiGen - devonfw Java Templates ................... SUCCESS [  1.015 s]
[INFO] CobiGen - Template Engine for language agnostic templates SUCCESS [  1.709 s]
...
```

CobiGen might be a complex product itself but it slightly looks as if it complex by design and could benefit from KISS principle ;)

hohwille:
I have spent more time to test my feature integratively with CobiGen in Eclipse than implementing the PR itself.
However, I found 2 problems this way:
1. the XML configuration was not correct and has now been fixed with my latest commit & push
2. CobiGen has this `variables.*` hack that was IMHO introduced due to the nature of freemarker and whatever legacy reasons. I expected the variables from the trigger to be also present as root variables in the map but this is not the case as I could figure out during debug session. So I also fixed this with my last commit & push.

Two problems are remaining:
* CobiGen is not a regular maven project so I was not aware that my dependency from agnostic-template-engine to templates-devonfw-java is "not allowed" and breaks the custom build and pipelines of CobiGen. For me it would be natural and intuitive to use my templates for agnostic engine also to test the engine implementation what works fine in JUnits and regular maven.
* My agnostic engine is using service loader to discover generators. The templates (e.g. devonfw-java) therefore can contain custom generator implementations and declare them using (legacy) service loader descriptors (`META-INF/services`). For whatever reason this does not work in my manual integration tests. So I assume that some hyper magic classloader stuff is preventing this. My assumption would simply be that during the generation the templates project with all its dependencies (as well as the project where the selected Input comes from with all its deps) is on the classpath and therefore service-loader will fine my custom generators. However, no generator is ever found - whatever I do (even when running in CCL).

For these remaining problems I see no other chance than that some CobiGen guru is taking some action and provides the last tiny piece to the puzzle to finally make it work and the vision come to action.
As this is my fourth PR for this story and I did not yet earn a single comment and any of them, I am slightly concerned that my story (#158) will need another 8 years to become reality...

